PR #61, opened 2023-03-08
Don't add default values to config map if not specified.

### LINKED ISSUE #53: Update Configmap template
Go templates support conditionals

https://github.com/project-codeflare/codeflare-operator/blob/main/config/internal/mcad/configmap.yaml.tmpl currently creates non-empty (but "falsy/null") keys, which doesn't match the behaviour of the upstream helm charts. We should update the templates to look similar.
https://github.com/project-codeflare/multi-cluster-app-dispatcher/blob/main/deployment/mcad-controller/templates/configmap.yaml

KPostOffice:
This allows MCAD to determine what an undefined value should default to

closes #53


anishasthana:
At first glance this looks good. I'll run through some tests with built images on Monday

PR #62, opened 2023-03-08
Adding Image Build/Push CI

Maxusmusti:
Adds core image and bundle image build/push to CI upon version change in push to main


anishasthana (commented):
/lgtm

Overall this looks good to me. How would we verify it?

KPostOffice:
Does the version need to be maintained in multiple places?

Maxusmusti:
> Does the version need to be maintained in multiple places?

I believe there are two other places already where the full operator version is specified, one being `config/manifests/bases/codeflare-operator.clusterserviceversion.yaml` and the other being `bundle/manifests/codeflare-operator.clusterserviceversion.yaml`

KPostOffice:
I personally don't think having the version in the code is a good idea. The code doesn't have a version until we create an actual release/tag in git. The workflow should basically be the same, but the created bundle should be added as an artifact to the release.

https://stackoverflow.com/questions/61891328/trigger-github-action-only-on-new-tags
https://trstringer.com/github-actions-create-release-upload-artifacts/

.github/workflows/image.yml
```diff
@@ -0,0 +1,51 @@
...
+      with:
+        username: ${{ secrets.RH_REG_ID }}
+        password: ${{ secrets.RH_REG_TOKEN }}
+        registry: registry.redhat.io
+
+    - name: Image Build
+      run: |
+        make image-build -e IMG=quay.io/project-codeflare/codeflare-operator:${SOURCE_TAG}
+        podman tag quay.io/project-codeflare/codeflare-operator:${SOURCE_TAG} quay.io/project-codeflare/codeflare-operator:latest
+        make bundle-build -e BUNDLE_IMG=quay.io/project-codeflare/codeflare-operator-bundle:${SOURCE_TAG}
```
KPostOffice, line 39: does `make bundle` need to be called? From what I can tell, this will just build an image with whatever bundle files already exist on the repo.

Maxusmusti: Oh hmm, yeah good catch

PR #58, opened 2023-03-08
Update instaScale crd and controller

tedhtchang:
Close #52

PR #63, opened 2023-03-09
Allow resource specification in MCAD CR

MichaelClifford:
This PR allows users to specify their resource requirements from within the MCAD CR yaml. 

Big thanks to @tedhtchang for leading the way with #58


api/v1alpha1/mcad_types.go
```diff
@@ -51,6 +52,9 @@ type MCADSpec struct {
 	PodCreationTimeout int `json:"podCreationTimeout,omitempty"`
 	//podCreationTimeout: //int (default blank)
 
+	// ControllerResources TODO: Add details
```
anishasthana, line 55: I think we could aadd details on this, something like "ControllerResources is responsible for determining what the CPU and Memory limits/requests" are

controllers/testdata/mcad_test_cases/case_1.yaml
```diff
@@ -2,4 +2,11 @@ apiVersion: codeflare.codeflare.dev/v1alpha1
 kind: MCAD
 metadata:
   name: blank-custom-resource
-spec: {}
+spec:
```
anishasthana, line 5: We should create a new test case that has the controller resources defined.

anishasthana (changes requested):
Thanks for the PR! Had a couple comments.

controllers/testdata/mcad_test_cases/case_1.yaml
MichaelClifford: @anishasthana we should keep this test case as is, and add a second that includes the resource definitions? Right?

MichaelClifford:
Thanks @anishasthana changes made :smile:

Maxusmusti (changes requested):
Rebase required, a couple other PRs merged

MichaelClifford:
@Maxusmusti Ok, rebased, but had to make a few additional changes to accommodate the other PR's. Please review again.

Maxusmusti (approved):
/lgtm

tested on a cluster with various configurations

Issue #36, opened 2023-03-01
Verify reconciliation of owner-less resources


anishasthana:
The CodeFlare operator is currently creating a number of cluster scoped objects as well as objects in namespaces other than the defining CR. It is possible that reconciliation of these objects (after actions such as deleting/editing) is not working as expected.

PR #72, opened 2023-03-15
add watches for all MCAD and InstaScale ownerless resources

KPostOffice:
created a handler that gets a specific MCAD cr from labels and enqueues its reconciliation on a change

closes: #36


Maxusmusti (changes requested):
`{name}-mcad-custom-metrics-server-admin-role` role seems to have been left out of the resources to get watched, can be deleted and only comes back when one of the other roles is deleted. Also, not sure if this is expected, but the MCAD deployment itself can be deleted without it getting reconciled and brought back up. Everything else for MCAD seemed to work as expected though, haven't tested instascale yet.

KPostOffice:
@Maxusmusti I've updated the PR, it should address this issue now.

Maxusmusti (approved):
/lgtm

Tested every component on a cluster, nothing could be deleted when CR existed, and everything was deleted when CR was removed, as expected

PR #90, opened 2023-04-21
Update default MCAD image

anishasthana:


tedhtchang (approved):
Verified the following way:
```
make manifest
make install
make run
cat <<EOF | oc apply -f -
apiVersion: codeflare.codeflare.dev/v1alpha1
kind: MCAD
metadata:
  name: example
  namespace: default
spec: {}
EOF
# make sure no error in the log
oc logs mcad-controller-example-64cb4c6fd-v8kvw 
```

/lgtm

anishasthana:
Merging manually for now as we have an lgtm + "approve" on the PR, just not from the right people. I will submit a follow-up PR soon to update the owners list.

PR #94, opened 2023-05-05
Add bundle unit action

anishasthana:
This PR does the following:
1. Split bundle validation out into a separate action
2. Introduce a new action that checks to see if any code changes occurred that should result in a new bundle. If the new bundle is not already included in the PR, the PR will fail.

Open questions:
1) I was unable to figure out how to add the replaces and createdAdd fields, so I chose to just ignore them in the git diff.
2) I _think_ we only need the make bundle in here. `make bundle` calls `make manifests`. The one you could debate is including `make generate`  as well.

For testing you can copy the make commands alongside the git diffs and run them yourselves manually.


KPostOffice:
What's the reason for having the bundle in the git repo, aren't Operator bundles basically build artifacts?

anishasthana:
> What's the reason for having the bundle in the git repo, aren't Operator bundles basically build artifacts?

I think it'll give us an easier way to track changes that have been happening to the operator over time. With the 0.0.2 release the bundle generation resulted in a ton of changes, and not all of them made sense to me.

I've definitely seen a number of other operators doing the same, so I don't think it's an anti-pattern either.

KPostOffice:
> > What's the reason for having the bundle in the git repo, aren't Operator bundles basically build artifacts?
> 
> I think it'll give us an easier way to track changes that have been happening to the operator over time. With the 0.0.2 release the bundle generation resulted in a ton of changes, and not all of them made sense to me.
> 
> I've definitely seen a number of other operators doing the same, so I don't think it's an anti-pattern either.

Can you link me to the operators? I am just wary of including the build artifacts in general. As for comparing bundles, a developer can just checkout 2 git refs and run a `diff` on the generated bundles.

I feel that generating new bundles for every pull request will just clutter the PR with code that wasn't actually authored, and it makes me question what it is actually tracking if it is all just generated by `operator-sdk generate bundle`.

Makefile
```diff
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.0.1
+VERSION ?= 0.0.2
```
tedhtchang, line 6: We should also add `  replaces: codeflare-operator.v0.0.1` the spec in the [csv template](https://github.com/project-codeflare/codeflare-operator/blob/main/config/manifests/bases/codeflare-operator.clusterserviceversion.yaml#L66).

tedhtchang (dismissed):

jbusche:
The `make bundle` worked for me on Mac, resulting with a clusterserviceversion in 
```
bundle/manifests/codeflare-operator.clusterserviceversion.yaml
```
To get it to build on my Red Hat 8, I changed the Makefile curl command slightly:
```
#curl -L $(OPERATOR_SDK_DL_URL)/operator-sdk_$(shell go env GOOS)_$(shell go env GOARCH) --output-dir $(LOCALBIN) --output operator-sdk
        curl -L $(OPERATOR_SDK_DL_URL)/operator-sdk_$(shell go env GOOS)_$(shell go env GOARCH) --output $(LOCALBIN)/operator-sdk
```
and then it builds.

The only thing weird to me is, I was expecting some kind of versioning folder(s) more similar to this:
https://github.com/redhat-openshift-ecosystem/community-operators-prod/tree/main/operators/codeflare-operator
Where the we'd see the version folders:
```
ls -lrt codeflare-operator/
drwxr-xr-x 5 root root  52 May 10 12:08 0.0.1
-rw-r--r-- 1 root root 191 May 10 12:08 ci.yaml
drwxr-xr-x 5 root root  52 May 10 12:08 0.0.2
```
It also doesn't know about the replaces, like you have in the community operator [here](https://github.com/redhat-openshift-ecosystem/community-operators-prod/blob/main/operators/codeflare-operator/0.0.2/manifests/codeflare-operator.clusterserviceversion.yaml#L556)

Makefile
tedhtchang: Maybe need something like:
```
# best if we can automatically detect the previous version
PREVIOUS_VERSION = 0.0.1
# then we can add a patch in the `bundle` target
cd config/manifests && $(KUSTOMIZE) edit add patch --patch '[{"op":"add", "path":"/spec/replaces", "value": "codeflare-operator.v$(PREVIOUS_VERSION)" }]' --kind ClusterServiceVersion
```

tedhtchang: Tracking this in a separate [issue](https://github.com/project-codeflare/codeflare-operator/issues/98)

tedhtchang (commented):
/lgtm

jbusche (approved):
Looks good, thanks!

PR #110, opened 2023-05-24
add new github action which tags and builds a new operator

### LINKED ISSUE #98: Add replaces to the base CSV template and patch the value with the previous version
Since we have multiple versions we have an update path in the CSV file. We need to:
1. Add spec.replaces in the [base CSV file](https://github.com/project-codeflare/codeflare-operator/blob/main/config/manifests/bases/codeflare-operator.clusterserviceversion.yaml#L66)
2. Add/detect the previous version (in the [Makefile](https://github.com/project-codeflare/codeflare-operator/blob/main/Makefile))
Patch the value  of spec.replaces with the previous version. For example:
```
# best if we could detect this. If we cannot, we need to document it somewhere.
PREVIOUS_VERSION = 0.0.1 
# then we can add a patch in the `PHONY: bundle`
cd config/manifests && $(KUSTOMIZE) edit add patch --patch '[{"op":"add", "path":"/spec/replaces", "value": "codeflare-operator.v$(PREVIOUS_VERSION)" }]' --kind ClusterServiceVersion
```
See [here](
https://github.com/project-codeflare/codeflare-operator/pull/94#discussion_r1190298877) for more detail

KPostOffice:
fixes #98


Makefile
```diff
@@ -3,6 +3,10 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
+# best if we could detect this. If we cannot, we need to document it somewhere.
+# then we can add a patch in the `PHONY: bundle`
+
+PREVIOUS_VERSION ?= 0.0.1
 VERSION ?= 0.0.2
```
tedhtchang, line 10: Ran `make bundle -e PREVIOUS_VERSION=0.0.3 -e VERSION=0.0.4`. No error.
To make the next release easier, should we preset the envs to:
```suggestion
PREVIOUS_VERSION ?= 0.0.3
VERSION ?= 0.0.4
```

.github/workflows/tag-and-build.yml
```diff
@@ -0,0 +1,76 @@
...
+            sha: context.sha
+          })
+
+    - name: Install operator-sdk
+      run: |
+        export ARCH=$(case $(uname -m) in x86_64) echo -n amd64 ;; aarch64) echo -n arm64 ;; *) echo -n $(uname -m) ;; esac)
+        export OS=$(uname | awk '{print tolower($0)}')
+        export OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/v1.27.0
+        curl -LO ${OPERATOR_SDK_DL_URL}/operator-sdk_${OS}_${ARCH}
+        chmod +x operator-sdk_${OS}_${ARCH} && sudo mv operator-sdk_${OS}_${ARCH} /usr/local/bin/operator-sdk
```
tedhtchang, line 46: Is there a reason not using `make install-operator-sdk` ?

tedhtchang (commented):
Hi @KPostOffice I don't know if I could test out the the git action because I don't have the credentials to the the Rh registry. Otherwise looks good just some nit picks.

.github/workflows/tag-and-build.yml
KPostOffice: I was using an existing git action. https://github.com/project-codeflare/codeflare-operator/blob/main/.github/workflows/operator-image.yml I can see if I can do this the way that you suggested

KPostOffice:
> Hi @KPostOffice I don't know if I could test out the the git action because I don't have the credentials to the the Rh registry. Otherwise looks good just some nit picks.

Haven't tested yet. I will try the changes on my fork.

Makefile
tedhtchang: I think `VERSION ?= 0.0.0-dev` should be ok. `PREVIOUS_VERSION` may need to be an existing bundle release judging from some examples [here](https://docs.openshift.com/container-platform/4.4/operators/understanding_olm/olm-understanding-olm.html#olm-upgrades_olm-understanding-olm) which implies the `replaces` is an version from an update graph which maybe built from  existing CSVs.

tedhtchang:
@KPostOffice I could not test the the new github workflow. The make bundle /lgtm.

anishasthana:
What do folks think of switching over https://github.com/project-codeflare/codeflare-operator/blob/main/Dockerfile#L2 to be one of the registry.access.redhat.com/ubi8/go-toolset:1.18.9-8? That oughta get around the need to authenticate to the red hat registry.

.github/workflows/tag-and-build.yml
```diff
@@ -0,0 +1,70 @@
...
+    - uses: actions/checkout@v3
+
+    - name: Activate cache
+      uses: actions/cache@v2
+      with:
+        path: /cache
+        key: ${{ runner.os }}-cache-${{ hashFiles('**/go.sum', '.pre-commit-config.yaml') }}
+
+    - name: Create tag
+      uses: actions/github-script@v5
```
sutaakar, line 29:
```suggestion
      uses: actions/github-script@v6
```
A nitpick: v5 is using deprecated Node 12.

KPostOffice:
> registry.access.redhat.com/ubi8/go-toolset:1.18.9-8

More than happy to do this. I can then also remove the auth step from the GitHub actions

jbusche (approved):
LGTM - Tested together with Kevin on my fork, the github actions worked beautifully and pushed the candidate release to my quay repo.

PR #119, opened 2023-06-01
Use uniquely identifying labels selector for the operator deployment

### LINKED ISSUE #112: The operator deployment should use non-overlapping label selector
The operator deployment is configured with the default label selector, as generated by the kubebuilder scaffolding, i.e.:  

```yaml
spec:
  selector:
    matchLabels:
      control-plane: controller-manager
```

This selector can intersect with other deployments, for instance the Open Data Hub operator, which leads to issues where selection based on this selector returns more than expected, e.g., in the OpenShift Web console:

Screenshot 2023-05-30 at 12 33 32

This should be changed, so it's disjunct from any other deployments, ideally following Kubernetes recommended labels:

https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

The main challenge is that deployment label selector is immutable once created, which will break the rollout deployment performed during the next OLM upgrade:

https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates

With some more background in kubernetes/kubernetes#50808.

As the project is in its early phase, the simplest approach would be to document in the release note manual deletion of the previous deployment is required.

astefanutti:
This PR changes the operator deployment to use a uniquely identifying labels selector.

In order to work-around the immutability of the labels selector field, the deployment has been renamed according to OLM recommendation: https://github.com/operator-framework/operator-lifecycle-manager/issues/952#issuecomment-639657949.

This PR also leverages Kustomise `commonLabels` config, to apply [Kubernetes recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) to all resources.

Fixes #112.

### DIFF
--- a/config/default/manager_config_patch.yaml
+++ b/config/default/manager_config_patch.yaml
@@ -1,7 +1,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: controller-manager
+  name: manager
   namespace: system
 spec:
   template:
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -9,8 +9,9 @@ namespace: openshift-operators
 namePrefix: codeflare-operator-
 
 # Labels to add to all resources and selectors.
-# commonLabels:
-#  someName: someValue
+commonLabels:
+  app.kubernetes.io/name: codeflare-operator
+  app.kubernetes.io/part-of: codeflare
 
 bases:
 - ../crd
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -3,7 +3,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: controller-manager
+  name: manager
   namespace: system
 spec:
   template:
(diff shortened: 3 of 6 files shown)


config/default/manager_config_patch.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: controller-manager
+  name: manager
```
KPostOffice, line 4: Should we also change some of these names to be more specific? `codeflare-manager` or something?

astefanutti: All the resource names are prefixed using `namePrefix: codeflare-operator-`, as per:

https://github.com/project-codeflare/codeflare-operator/blob/d03836b06599da328684e53cd6ba071eeb6b3e0e/config/default/kustomization.yaml#L9

so the name rendered by Kustomize here is `codeflare-operator-manager`.

KPostOffice: Ah, missed this. Thanks

KPostOffice: this /lgtm then

KPostOffice: this /lgtm then

astefanutti: Thanks

anishasthana (approved):
/lgtm

Thanks @astefanutti

PR #125, opened 2023-06-07
Remove generated OLM bundle Dockerfile

astefanutti:
This PR removes the `bundle.Dockerfile` file, that's generated by the Operator SDK. It also makes sure the bundle is generated prior to building the bundle container image.


anishasthana (approved):
/lgtm

PR #127, opened 2023-06-08
Update operator to 0.0.4

anishasthana:
One thing to note: I haven't yet tested with the latest images

### DIFF
--- a/config/internal/mcad/deployment.yaml.tmpl
+++ b/config/internal/mcad/deployment.yaml.tmpl
@@ -25,7 +25,7 @@ spec:
           envFrom:
             - configMapRef:
                 name: mcad-{{.Name}}-config
-          image: 'quay.io/project-codeflare/mcad-controller:release-v1.29.57'
+          image: 'quay.io/project-codeflare/mcad-controller:release-v1.31.0'
           imagePullPolicy: Always
           ports:
             - name: https
--- a/controllers/defaults.go
+++ b/controllers/defaults.go
@@ -5,5 +5,5 @@ package controllers
 // ***********************
 
 const (
-	InstaScaleImage = "quay.io/project-codeflare/instascale-controller:v0.0.3"
+	InstaScaleImage = "quay.io/project-codeflare/instascale-controller:v0.0.4"
 )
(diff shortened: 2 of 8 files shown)


controllers/defaults.go
```diff
@@ -5,5 +5,5 @@ package controllers
 // ***********************
 
 const (
-	InstaScaleImage = "quay.io/project-codeflare/instascale-controller:v0.0.3"
+	InstaScaleImage = "quay.io/project-codeflare/instascale-controller:v0.0.4"
```
anishasthana, line 8: @astefanutti if we don't edit this file how else would we update the default?

astefanutti: It's now to be set in the Makefile: https://github.com/project-codeflare/codeflare-operator/blob/a71303b7d9a2a59eb8310dbbaa351e2b43af256c/Makefile#L13

Ideally, we'll eventually have the dependencies defined there in a central place.

astefanutti: And run `make defaults`, but it's done with most make targets.

config/internal/mcad/deployment.yaml.tmpl
```diff
@@ -25,7 +25,7 @@ spec:
           envFrom:
             - configMapRef:
                 name: mcad-{{.Name}}-config
-          image: 'quay.io/project-codeflare/mcad-controller:release-v1.29.57'
+          image: 'quay.io/project-codeflare/mcad-controller:release-v1.31.0'
```
astefanutti, line 28: This has to be update here for the MCAD CRDs: 

https://github.com/project-codeflare/codeflare-operator/blob/a71303b7d9a2a59eb8310dbbaa351e2b43af256c/config/crd/mcad/kustomization.yaml#L2

controllers/defaults.go
anishasthana: Got it!

config/internal/mcad/deployment.yaml.tmpl
anishasthana: Updated. Good catch!

controllers/defaults.go
anishasthana: Fixed.

anishasthana:
I opened #128 for the test cases fix.

anishasthana:
/retest

The openshift ci failure seems to be a flake on their end.

PR #131, opened 2023-06-09
v0.0.4 bundle generated stuff

anishasthana:


config/manifests/kustomization.yaml
```diff
@@ -14,3 +14,9 @@ patches:
 - patch: '[{"op":"add", "path":"/metadata/annotations/containerImage", "value": "quay.io/project-codeflare/codeflare-operator:v0.0.2" }]'
   target:
     kind: ClusterServiceVersion
+- patch: '[{"op":"add", "path":"/metadata/annotations/containerImage", "value": "quay.io/project-codeflare/codeflare-operator:v0.0.4" }]'
```
astefanutti, line 17: I think this can actually be removed, as the patch is done in the Makefile:

https://github.com/project-codeflare/codeflare-operator/blob/95a99358ed3c60a0ef9a814a6bec67647be85a39/Makefile#L240-L241

config/manager/kustomization.yaml
```diff
@@ -13,4 +13,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: quay.io/project-codeflare/codeflare-operator
-  newTag: v0.0.2
+  newTag: v0.0.4
```
astefanutti, line 16: One alternative would be to removed the hard-coded value, and patch the configuration from the Makefile.

PR #136, opened 2023-06-13
Initial e2e tests

### LINKED ISSUE #4: Set up e2e tests for codeflare stack

astefanutti:
This PR adds e2e tests for the CodeFlare stack, more specifically, MCAD, KubeRay/Ray and the CodeFlare SDK.

It includes:

- [x] A e2e test suite with 3 initial tests:
  - [x] MNIST training submitted as a RayJob to a RayCluster managed by MCAD
  - [x] MNIST training submitted as a batch Job managed by MCAD
  - [x] MNIST training on a Ray cluster via the CodeFlare SDK 
  (only runs on OpenShift, requires project-codeflare/codeflare-sdk#146)
- [x] A GitHub Actions workflow, that runs e2e tests against a KinD test cluster.
- [x] The parameterisation of the components under tests, so the test suite can be reused by downstream projects. 
- [x] The documentation on how to run / debug the e2e tests locally. 

The execution of the e2e workflow currently takes around 14 minutes on average. A third of that execution time corresponds to the build of the operator container image, which may be optimised in the future, by caching layers for example. 

Fixes #4.


.github/workflows/e2e_tests.yaml
```diff
@@ -0,0 +1,182 @@
...
+          echo "$(pwd)/bin" >> $GITHUB_PATH
+
+      - name: Set Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: v1.18
+
+      - name: Container image registry
+        run: |
+          podman run -d --restart=always -p 5001:5000 --name registry registry:2.8.1
```
jiripetrlik, line 73: - Thinking if "--restart=always" is necessary in Github action. It will run only once without any restarts.
- Is port 5000 already in use on Github machines? Can we use standard port for Docker registry: 5000:5000?

.github/workflows/e2e_tests.yaml
```diff
@@ -0,0 +1,182 @@
...
+  group: ${{ github.head_ref }}-${{ github.workflow }}
+  cancel-in-progress: true
+
+jobs:
+  kubernetes:
+
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Cleanup
```
jiripetrlik, line 33: Is this clean up necessary? I think that Github actions starts with clean VM.

.github/workflows/e2e_tests.yaml
```diff
@@ -0,0 +1,182 @@
...
+      - name: Set Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: v1.18
+
+      - name: Container image registry
+        run: |
+          podman run -d --restart=always -p 5001:5000 --name registry registry:2.8.1
+
+          export REGISTRY_ADDRESS=$(hostname -i):5001
```
jiripetrlik, line 75: Maybe "localhost" or "127.0.0.1" instead of hostname command will also work. I think that on some different machine it may return multiple IP addresses in the output which may break the functionality.

.github/workflows/e2e_tests.yaml
astefanutti: Good points, I've removed these.

.github/workflows/e2e_tests.yaml
astefanutti: It frees some GBs of disk space, which might be useful. I'll double check what's the pressure on the file system.

.github/workflows/e2e_tests.yaml
astefanutti: I remember I tried it and it failed. If I remember it correctly, the root cause is connected to IPv6 support in Kubernetes / KinD. Somehow, with `localhost` or even `127.0.0.1`, it involves an IPv6 loopback interface, which breaks the pulling of the image in KinD.

I've just retried it, and it indeed fails: https://github.com/project-codeflare/codeflare-operator/actions/runs/5289718123/jobs/9572984628?pr=136#step:9:229

jiripetrlik: Yes, it is true. Local registry will not be visible from Kind cluster under the IP: 127.0.0.1. It is true. Maybe I have one more idea. We can try to get the IP from docker network inspect kind | jq -r '.[0].IPAM.Config[0].Gateway' . But it is probably a detail.

.github/workflows/e2e_tests.yaml
```diff
@@ -0,0 +1,190 @@
...
+      - 'LICENSE'
+
+concurrency:
+  group: ${{ github.head_ref }}-${{ github.workflow }}
+  cancel-in-progress: true
+
+jobs:
+  kubernetes:
+
+    runs-on: ubuntu-20.04
```
sutaakar, line 30: Is there any reason to use specific version and not `ubuntu-latest`?

astefanutti: I've experienced workflows failing over night when using rolling tags, and found it sometimes difficult to realise the root cause was the implicit upgrade of a dependency. I think build / test run reproducibility largely overweights the cost of explicit upgrades, at least in the case of an OS upgrade.

README.md
```diff
@@ -14,7 +15,36 @@ CodeFlare Stack Compatibility Matrix
...
+1. Setup the test cluster:
+
+    ```bash
+    # Create a KinD cluster
+    $ kind create cluster --image kindest/node:v1.25.8
+    # Install the CRDs
+    $ make install
+    ```
+
+2. Start the operator locally:
```
sutaakar, line 33: And what about starting MCAD (https://github.com/project-codeflare/codeflare-operator/blob/ce9349577536ed59d2f66c7b57550f64a79b74a7/.github/workflows/e2e_tests.yaml#L116-L164)?
Isn't it also required for test execution?

sutaakar: And Kuberay?

README.md
```diff
@@ -14,7 +15,36 @@ CodeFlare Stack Compatibility Matrix
...
+
+The e2e tests can be executed locally by running the following commands:
+
+1. Setup the test cluster:
+
+    ```bash
+    # Create a KinD cluster
+    $ kind create cluster --image kindest/node:v1.25.8
+    # Install the CRDs
+    $ make install
```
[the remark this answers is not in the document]
astefanutti: This will be fixed with #152.

README.md
astefanutti: Ah right, these two steps are missing 👍🏼.

test/e2e/mnist_pytorch_mcad_job_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	test.T().Logf("Waiting for Job %s/%s to complete", job.Namespace, job.Name)
+	test.Eventually(Job(test, job.Namespace, job.Name), TestTimeoutLong).Should(
+		Or(
+			WithTransform(ConditionStatus(batchv1.JobComplete), Equal(corev1.ConditionTrue)),
+			WithTransform(ConditionStatus(batchv1.JobFailed), Equal(corev1.ConditionTrue)),
+		))
+
+	// Print the job logs
+	PrintJobLogs(test, job.Namespace, job.Name)
```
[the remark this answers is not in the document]
astefanutti: I used to defer the log printing exactly as you suggested, but it'll have to wait until the successful assertion timed out. So I changed the assertion to wait for completion, whether failed or successful, and always print the logs. That way it fails as soon as the job complete.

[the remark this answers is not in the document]
astefanutti: Great idea for storing the job logs in file, and maybe use https://github.com/actions/upload-artifact to store them as part of the workflow execution.

test/e2e/mnist_rayjob_mcad_raycluster_test.go
```diff
@@ -0,0 +1,262 @@
...
+			ShutdownAfterJobFinishes: false,
+		},
+	}
+	rayJob, err = test.Client().Ray().RayV1alpha1().RayJobs(namespace.Name).Create(test.Ctx(), rayJob, metav1.CreateOptions{})
+	test.Expect(err).NotTo(HaveOccurred())
+	test.T().Logf("Created RayJob %s/%s successfully", rayJob.Namespace, rayJob.Name)
+
+	test.T().Logf("Waiting for RayJob %s/%s to complete successfully", rayJob.Namespace, rayJob.Name)
+	test.Eventually(RayJob(test, namespace, rayJob.Name), TestTimeoutLong).
+		Should(WithTransform(RayJobStatus, Equal(rayv1alpha1.JobStatusSucceeded)))
```
sutaakar, line 255: Can you add the check for succeeded or failed job also here?

sutaakar:
@astefanutti Would it have sense to add comments to the tests describing what functionality is tested in specific tests? Something like a description from [test strategy ADR](https://github.com/project-codeflare/adr/pull/6).

test/e2e/mnist_rayjob_mcad_raycluster_test.go
astefanutti: Good catch. Done with e29c28d8e2b36ea8808cca15f242a3ae7741da3b.

astefanutti:
> @astefanutti Would it have sense to add comments to the tests describing what functionality is tested in specific tests? Something like a description from [test strategy ADR](https://github.com/project-codeflare/adr/pull/6).

@sutaakar yes, absolutely. We can start with commenting the test case functions, and research ways to structure more that information in the next cycles.

KPostOffice (commented):
Some questions. I'm unfamiliar with golang testing.

.github/workflows/e2e_tests.yaml
```diff
@@ -0,0 +1,190 @@
...
+
+          mkdir -p "$(pwd)/bin"
+          echo "$(pwd)/bin" >> $GITHUB_PATH
+
+      - name: Set Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: v1.18
+
+      - name: Container image registry
```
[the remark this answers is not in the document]
sutaakar: From my perspective this is faster, doesn't require Quay credentials and doesn't spam registry with hundreds of images unusable outside of related PR.

.pre-commit-config.yaml
```diff
@@ -25,5 +25,4 @@ repos:
     hooks:
       - id: go-fmt
       - id: golangci-lint
-      - id: go-build
```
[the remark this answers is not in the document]
astefanutti: Actually I think we should migrate to calling make targets directly, as https://github.com/dnephin/pre-commit-golang is not maintained anymore.

test/support/utils.go
```diff
@@ -0,0 +1,21 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package support
+
+func Ptr[T any](v T) *T {
```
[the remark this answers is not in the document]
astefanutti: The main reason is that it's not generic. There are cases where it's needed to get a pointer of a struct, inlined. That package only provides it for primitive types and `time.Duration`.

test/support/client.go
```diff
@@ -0,0 +1,99 @@
...
+
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+
+	codeflareclient "github.com/project-codeflare/codeflare-operator/client/clientset/versioned"
+	mcadclient "github.com/project-codeflare/multi-cluster-app-dispatcher/pkg/client/clientset/controller-versioned"
+	rayclient "github.com/ray-project/kuberay/ray-operator/pkg/client/clientset/versioned"
+)
+
+type Client interface {
```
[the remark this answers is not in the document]
astefanutti: Generally clients are defined as interfaces, so different implementations can be provided, for e.g., caching, mocking, instrumenting, purposes.

test/e2e/mnist_pytorch_mcad_job_test.go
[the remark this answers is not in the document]
astefanutti: Done with ba6916ceac40431750f7ccf60ddca052351519d6.

.github/workflows/e2e_tests.yaml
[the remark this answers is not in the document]
KPostOffice: Okay, that's fine by me. If someone wants to test with the PR image they can just build it themselves.

test/support/client.go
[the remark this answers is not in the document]
KPostOffice: I think my main confusion is that it is defined only inside the `test` dir so I wasn't sure when we would create other implementations for it.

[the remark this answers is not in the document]
astefanutti: You're right, there is only one. I've just followed the idiomatic pattern, but it could be replaced with the implementing `struct` here.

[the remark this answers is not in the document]
KPostOffice: ++, makes sense. Happy to follow patterns

.pre-commit-config.yaml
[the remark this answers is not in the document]
KPostOffice: https://github.com/project-codeflare/codeflare-operator/issues/156

README.md
astefanutti: Done with ff32292c1eba52174a5eb248f4fa56144c79aba8.

astefanutti:
> > @astefanutti Would it have sense to add comments to the tests describing what functionality is tested in specific tests? Something like a description from [test strategy ADR](https://github.com/project-codeflare/adr/pull/6).
> 
> @sutaakar yes, absolutely. We can start with commenting the test case functions, and research ways to structure more that information in the next cycles.

Done with 2045c7ad31f7feb77dff35b93a666188b8e93e87.

sutaakar (approved):
Looks good to me

anishasthana (approved):

README.md
```diff
@@ -14,7 +15,43 @@ CodeFlare Stack Compatibility Matrix
...
+2. Set up the CodeFlare stack:
+   ```bash
+   $ make setup-e2e
+   ```
+
+3. Start the operator locally:
+
+    ```bash
+    $ make run
+    ```
```
KPostOffice, line 42: I think these should be swapped. Make setup-e2e assumes that the operator is running so that it can create an MCAD deployment

astefanutti: Ah right, or the `kubectl wait` command that waits for the MCAD deployment to become ready can be removed. It's not strictly necessary.

astefanutti: Done with #163.

astefanutti: Thanks!

PR #139, opened 2023-06-14
Add MCAD version variable

astefanutti:
This PR adds a variable to parameterise the version of the MCAD dependency.

It follows up #116, so the version used to retrieve the MCAD CRDs is not hard-coded anymore.

It'll also be used for #114.

### DIFF
--- a/config/crd/mcad/kustomization.yaml
+++ b/config/crd/mcad/kustomization.yaml
@@ -1,2 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
 resources:
-  - github.com/project-codeflare/multi-cluster-app-dispatcher/config/crd?ref=release-v1.31.0
+- github.com/project-codeflare/multi-cluster-app-dispatcher/config/crd?ref=release-v0.0.0  # kpt-set: github.com/project-codeflare/multi-cluster-app-dispatcher/config/crd?ref=release-${MCAD_VERSION}
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,9 @@ VERSION ?= 0.0.0-dev
 # INSTASCALE_VERSION defines the default version of the InstaScale controller
 INSTASCALE_VERSION ?= v0.0.4
 
+# MCAD_VERSION defines the default version of the MCAD controller
+MCAD_VERSION ?= v1.31.0
+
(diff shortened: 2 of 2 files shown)


anishasthana (commented):
Should we also update https://github.com/project-codeflare/codeflare-operator/blob/main/CONTRIBUTING.md?plain=1#L34 as part of this?

@sutaakar fyi^^

sutaakar:
Would it be possible to update also https://github.com/project-codeflare/codeflare-operator/blob/main/config/internal/mcad/deployment.yaml.tmpl#L28 ?

astefanutti:
> Would it be possible to update also https://github.com/project-codeflare/codeflare-operator/blob/main/config/internal/mcad/deployment.yaml.tmpl#L28 ?

I've intentionally omitted it here, as it's the purpose of #114, that'll be similar to #120.

sutaakar (approved):
Looks good to me

config/crd/mcad/kustomization.yaml
```diff
@@ -1,2 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
 resources:
-  - github.com/project-codeflare/multi-cluster-app-dispatcher/config/crd?ref=release-v1.31.0
+- github.com/project-codeflare/multi-cluster-app-dispatcher/config/crd?ref=release-v0.0.0  # kpt-set: github.com/project-codeflare/multi-cluster-app-dispatcher/config/crd?ref=release-${MCAD_VERSION}
```
KPostOffice, line 4: do you think it's worth adding something like:
```
MCAD_REF ?= release-${MCAD_VERSION}
MCAD_REPO ?= github.com/project-codeflare/multi-cluster-app-dispatcher
MCAD_CRD ?= ${MCAD_REPO}/config/crd?ref=${MCAD_REF}
```
```suggestion
- github.com/project-codeflare/multi-cluster-app-dispatcher/config/crd?ref=release-v0.0.0  # kpt-set: ${MCAD_CRD}
```
or would that be overkill

anishasthana (approved):
/lgtm

config/crd/mcad/kustomization.yaml
astefanutti: Good idea, I've created #141 to follow up on this. PTAL.

PR #140, opened 2023-06-15
Updating rbacs for instascale

Fiona-Waters:
Relating to this instascale [PR](https://github.com/project-codeflare/instascale/pull/83)
and this github [issue](https://github.com/project-codeflare/instascale/issues/81).

### DIFF
--- a/config/internal/instascale/clusterrole.yaml.tmpl
+++ b/config/internal/instascale/clusterrole.yaml.tmpl
@@ -24,11 +24,19 @@ rules:
 - apiGroups:
   - ""
   resourceNames:
-    - instascale-ocm-secret
+  - instascale-ocm-secret
   resources:
-    - secrets
+  - secrets
   verbs:
-    - get
+  - get
+
+- apiGroups:
+  - config.openshift.io
+  resources:
+  - clusterversions
+  verbs:
+  - get
+  - list
 
 - apiGroups:
   - apps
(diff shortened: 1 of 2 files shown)


config/internal/instascale/clusterrole.yaml.tmpl
```diff
@@ -21,6 +21,13 @@ rules:
   - delete
   - patch
 
+- apiGroups:
+  - ""
+  resources:
+    - secrets
+  verbs:
+    - get
```
astefanutti, line 29: Could this be narrowed to specific resource names? Secrets are rather sensitive resources, so that's a best practice to avoid granting cluster-wide access whenever possible.

astefanutti: That'd be perfect 👍🏼.

Fiona-Waters: I will update. Thank you!

config/internal/instascale/clusterrole.yaml.tmpl
```diff
@@ -21,6 +21,23 @@ rules:
   - delete
   - patch
 
+- apiGroups:
```
astefanutti, line 24: I think that's an acceptable solution to fix the issue in the short term. However, I think the management of the OCM secret should be reviewed. I think it should be exposed in the InstaScale API as an object reference, and the RBAC configured accordingly. If that makes sense, could you please open an issue?

Fiona-Waters: Issue created and left in backlog. Thank you. https://github.com/project-codeflare/instascale/issues/87

PR #144, opened 2023-06-19
perf: addition of liveness and readiness probes to mcad deployment

dimakis:
[linked here in MCAD](https://github.com/project-codeflare/multi-cluster-app-dispatcher/pull/414)


astefanutti:
/hold

This requires an MCAD version upgrade.

config/internal/mcad/deployment.yaml.tmpl
```diff
@@ -40,6 +40,20 @@ spec:
...
+              path: /healthz
+              port: 8081
+            initialDelaySeconds: 30
+            timeoutSeconds: 5
+            periodSeconds: 5
+          readinessProbe:
+            httpGet:
+              path: /readyz
+              port: 8081
+            initialDelaySeconds: 30
```
sutaakar, line 54: Isn't 30 seconds too much?
How fast does MCAD usually start?

sutaakar: I am thinking whether there is actually any need for `initialDelaySeconds` for MCAD.
IMHO this is mainly leveraged for slowly starting containers when there is no need to run readiness/liveness probes too soon.
The image pulling doesn't count to the startup time, it should start once pod is provisioned.

dimakis: so we're in agreement here? there is no need for the `initialDelaySeconds` ?

astefanutti: +1

Issue #188, opened 2023-07-20
Unstable OLM Install and Upgrade PR check


sutaakar:
OLM Install and Upgrade PR check is unstable.
Based on a Subscription status it seems that the CodeFlare operator installation gets stuck in case any catalog is not properly started when Subscription is created.

PR #189, opened 2023-07-20
Remove waiting loops for kubernetes resources from Makefile

sutaakar:
Fixes https://github.com/project-codeflare/codeflare-operator/issues/188

### DIFF
--- a/.github/workflows/olm_tests.yaml
+++ b/.github/workflows/olm_tests.yaml
@@ -72,6 +72,11 @@ jobs:
           # wait for a while to be sure CRDs are installed
           sleep 1
           kubectl create -f https://github.com/operator-framework/operator-lifecycle-manager/releases/download/${OLM_VERSION}/olm.yaml
+          echo Wait for default CatalogSource to start
+          kubectl wait -n ${{ env.CATALOG_SOURCE_NAMESPACE }} catalogsource/${{ env.CATALOG_SOURCE_NAME }} --for=jsonpath='{.status.connectionState.lastObservedState}'=READY --timeout=180s
+        env:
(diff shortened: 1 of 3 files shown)


.github/workflows/olm_tests.yaml
```diff
@@ -73,6 +73,13 @@ jobs:
           sleep 1
           kubectl create -f https://github.com/operator-framework/operator-lifecycle-manager/releases/download/${OLM_VERSION}/olm.yaml
 
+      - name: Wait for default CatalogSource to start
+        run: |
+          make wait-for-catalog-source
```
astefanutti, line 78: Do you feel strong about defining the wait logic in the Makefile? I understand it may be to factorize the logic, but it's really a one-liner, and would make the test workflow easier to read.

Also I think `kubectl wait` supports `--for=jsonpath`, so instead of:

```
$ timeout 120 bash -c 'while [[ "$$(kubectl get catalogsource/'$(CATALOG_SOURCE_NAME)' -n '$(CATALOG_SOURCE_NAMESPACE)' -o json | jq -r .status.connectionState.lastObservedState)" != "READY" ]]; do sleep 5 && echo "$$(kubectl get catalogsource/'$(CATALOG_SOURCE_NAME)' -n '$(CATALOG_SOURCE_NAMESPACE)' -o json | jq -r .status.connectionState.lastObservedState)" ; done'
```

This could be:

```
$ kubectl wait -n ${CATALOG_SOURCE_NAMESPACE} catalogsource/'${CATALOG_SOURCE_NAME} --for=jsonpath='{.status.connectionState.lastObservedState}'=READY 
```

sutaakar: Good point, will refactor.

PR #192, opened 2023-07-23
Update: remove duplicated resource and should have lower in order

zdtsw:
https://github.com/project-codeflare/codeflare-operator/blob/main/controllers/mcad_controller.go#L232 has added resource


anishasthana (approved):
/lgtm

PR #193, opened 2023-07-24
Remove kube-rbac-proxy and open metrics endpoint

### LINKED ISSUE #191: Remove kube-rbac-proxy
## Description
`kube-rbac-proxy` was included as initial code as part of the Operator SDK framework. This proxy is adding more complexity towards exposing metrics.

To solve this issue, we made this PR https://github.com/project-codeflare/codeflare-operator/pull/185 which is bypassing the kube-rbac-proxy and exposing the CodeFlare Operator metrics successfully. However since we won't be making use of `kube-rbac-proxy`, it is best to remove it.

This was carefully discussed with @anishasthana, as well taking into consideration that the odh-operator repo is removing the proxy as well.

ChristianZaccaria:
Closes #191 #168


config/default/metrics_service.yaml
```diff
@@ -5,10 +5,10 @@ metadata:
   namespace: system
 spec:
   ports:
-  - name: https
+  - name: metrics
     port: 8443
```
sutaakar, line 9: I think this can be changed to 8080 too

ChristianZaccaria: You're right, seems to be more ideal as the targetPort is essentially 8080 too. Thanks @sutaakar

config/default/kustomization.yaml
```diff
@@ -25,11 +25,15 @@ bases:
...
+resources:
+# Add metrics service
+- metrics_service.yaml
+
 patchesStrategicMerge:
 # Protect the /metrics endpoint by putting it behind auth.
 # If you want your controller-manager to expose the /metrics
 # endpoint w/o any authn/z, please comment the following line.
-- manager_auth_proxy_patch.yaml
+# - manager_auth_proxy_patch.yaml
```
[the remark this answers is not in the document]
ChristianZaccaria: Excellent, you're right @anishasthana. I just made the last few changes there. The e2e tests should be passing now; after that I think it is good to merge.

anishasthana (commented):
/lgtm

controllers/mcad_controller.go
```diff
@@ -124,6 +124,8 @@ func (r *MCADReconciler) DeleteResource(params *MCADParams, template string, fns
 // +kubebuilder:rbac:groups=extensions,resources=replicasets,verbs=get;list;watch
 // +kubebuilder:rbac:groups=policy,resources=poddisruptionbudgets,verbs=get;list;watch
 // +kubebuilder:rbac:groups=storage.k8s.io,resources=csidrivers;csinodes;csistoragecapacities,verbs=get;list;watch
+// +kubebuilder:rbac:groups=authentication.k8s.io,resources=tokenreviews,verbs=create
```
anishasthana, line 127: Actually, why do we need this?

ChristianZaccaria: I'm not 100% sure, but this was brought up to me by @sutaakar as the e2e tests were failing when manually adding the tokenreviews, but when using kubebuilder to generate in the file instead, the e2e tests pass...

anishasthana: I should clarify, why do we need tokenreviews at all?

ChristianZaccaria: Previously I was getting this error: 
```
2023-07-26T09:16:08Z	INFO	Creating	{"name": "/mcad-mcad-controller-kube-scheduler-crb", "type": "rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding"}
2023-07-26T09:16:08Z	ERROR	Reconciler error	{"controller": "mcad", "controllerGroup": "codeflare.codeflare.dev", "controllerKind": "MCAD", "MCAD": {"name":"mcad","namespace":"codeflare-system"}, "namespace": "codeflare-system", "name": "mcad", "reconcileID": "4c657377-3718-4245-bd49-fa82bb5da2c6", 
"error": "clusterrolebindings.rbac.authorization.k8s.io \"mcad-mcad-controller-kube-scheduler-crb\" is forbidden: user \"system:serviceaccount:openshift-operators:codeflare-operator-controller-manager\" 
(groups=[\"system:serviceaccounts\" \"system:serviceaccounts:openshift-operators\" \"system:authenticated\"]) is attempting to grant RBAC permissions not currently held:\n{APIGroups:[\"authentication.k8s.io\"], Resources:[\"tokenreviews\"], Verbs:[\"create\"]}\n{APIGroups:[\"authorization.k8s.io\"], Resources:[\"subjectaccessreviews\"], Verbs:[\"create\"]}"} 
```
My understanding is that it seems that the Operator needed permission to create tokenreviews as it performs identity verification. In this case, the Operator receives requests from Prometheus when scraping metrics. [ServiceMonitor with bearerTokenFile](https://github.com/project-codeflare/codeflare-operator/blob/ca3baf682b6174e0a6424c1fae6734071237630a/config/prometheus/monitor.yaml#L12)

I suppose if we don't want to keep this, I could remove it? Please let me know what would be best to do

anishasthana (approved):

PR #197, opened 2023-07-27
Add Ray cluster REST API support in test support package

sutaakar:
# Issue link
Part of https://github.com/opendatahub-io/distributed-workloads/issues/63

# What changes have been made
Adjustments in e2e test support package to provide support for new ODH integration test - https://github.com/opendatahub-io/distributed-workloads/pull/71

# Verification steps
The changes can be verified by running ODH integration test, see instructions in https://github.com/opendatahub-io/distributed-workloads/pull/71

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [x] Testing is not required for this change

It is enough to run the local e2e tests (part of the PR check) to make sure nothing is broken.


sutaakar:
seems that k8s api upgrade is too big change for a codebase, need to find compatible OpenShift client versions (dependency hell strikes back)

test/support/ray_api.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"github.com/onsi/gomega"
+)
+
+func GetRayJobAPIDetails(t Test, rayClient RayClusterClient, jobID string) *RayJobDetailsResponse {
+	t.T().Helper()
+	return RayJobAPIDetails(t, rayClient, jobID)(t)
+}
+
+func WriteRayJobAPILogs(t Test, rayClient RayClusterClient, jobID string) {
```
astefanutti, line 28: I'd suggest to remove the `GetRayJobLogs` and `WriteRayJobLogs` functions from the `ray.go` file, that rely on the API server proxy, and have them replaced by these one, so there are used in the tests upstream as well.

For it to work, we can install the NGINX ingress controller in KinD (The KinD configuration already contains the "ingress-ready=true" kubelet extra argument), and create an Ingress resource in the `TestMNISTRayJobMCADRayCluster` test.

sutaakar: @astefanutti Created https://github.com/project-codeflare/codeflare-operator/issues/199

PR #198, opened 2023-07-27
Add verification of go import organization

anishasthana:


.github/workflows/verify_imports.yml
```diff
@@ -0,0 +1,17 @@
+name: Verify organization of imports
+on:
```
astefanutti, line 2: We may want to restrict it to Go files with `paths: - '**.go'`.

Makefile
```diff
@@ -440,3 +446,11 @@ test-e2e: defaults manifests generate fmt vet ## Run e2e tests.
...
 setup-e2e: ## Set up e2e tests.
 	KUBERAY_VERSION=$(KUBERAY_VERSION) test/e2e/setup.sh
+
+.PHONY: imports
+imports: openshift-goimports ## Organize imports in go files using openshift-goimports. Example: make imports
+	$(OPENSHIFT-GOIMPORTS)
+
+.PHONY: verify-imports
+verify-imports: openshift-goimports ## Run import verifications.
+	./hack/verify-imports.sh $(OPENSHIFT-GOIMPORTS)
```
astefanutti, line 456: We can inline the logic here if we want to.

client/clientset/versioned/scheme/register.go
```diff
@@ -19,12 +19,13 @@ limitations under the License.
...
 
 import (
-	codeflarev1alpha1 "github.com/project-codeflare/codeflare-operator/api/codeflare/v1alpha1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	runtime "k8s.io/apimachinery/pkg/runtime"
 	schema "k8s.io/apimachinery/pkg/runtime/schema"
 	serializer "k8s.io/apimachinery/pkg/runtime/serializer"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+
+	codeflarev1alpha1 "github.com/project-codeflare/codeflare-operator/api/codeflare/v1alpha1"
```
astefanutti, line 28: There are generated files. One option is to run openshift-goimports right after the client generation. Another option would be to exclude those files.

Makefile
anishasthana: I wasn't able to inline the logic in a way that looked clean, so I'd prefer to leave it in the script for now.

astefanutti: Sounds good, not a big deal.

PR #204, opened 2023-07-31
Add client action

anishasthana:
This PR adds github action checks for the following:
1) Generated client
2) Generated manifests
3) Generated code.

This should help ensure that our artifacts are always up to date.


astefanutti (commented):
Do you feel strong about splitting everything into its own workflow. It seems grouping the those checks, together with the openshift-goimports would make their execution faster and spare resources.

anishasthana:
I'll work on consolidating into one action.

anishasthana:
Consolidated.

astefanutti:
/lgtm

We could also add a verification step that check `go mod tidy` induces no diff.

tedhtchang (commented):
/LGTM tried [here](https://github.com/tedhtchang/codeflare-operator/actions/runs/5755502568)

PR #208, opened 2023-08-03
e2e: Use Ray dashboard API to retrieve job logs

### LINKED ISSUE #199: E2E tests - Retrieve Ray job logs using REST API only
Right now Ray job logs for e2e tests are retrieved by using two options:
- Using the [API server proxy](https://github.com/project-codeflare/codeflare-operator/blob/a82274750112f5d01b18f3de029549a720f12320/test/support/ray.go#L47)
- [Querying the exposed Ray cluster endpoint](https://github.com/project-codeflare/codeflare-operator/blob/a82274750112f5d01b18f3de029549a720f12320/test/support/ray_api.go#L28)

Purpose of this task is to consolidate the log retrieval to use exposed Ray cluster endpoint only.
For details see https://github.com/project-codeflare/codeflare-operator/pull/197#discussion_r1276346766

astefanutti:
Closes #199.

### DIFF
--- a/.github/actions/kind/action.yml
+++ b/.github/actions/kind/action.yml
@@ -47,3 +47,12 @@ runs:
         echo "KinD cluster:"
         kubectl cluster-info
         kubectl describe nodes
+
+    - name: Install Ingress controller
+      shell: bash
+      run: |
+        VERSION=controller-v1.6.4
+        echo "Deploying Ingress controller into KinD cluster"
+        curl https://raw.githubusercontent.com/kubernetes/ingress-nginx/"${VERSION}"/deploy/static/provider/kind/deploy.yaml | sed "s/--publish-status-address=localhost/--report-node-internal-ip-address\\n        - --status-update-interval=10/g" | kubectl apply -f -
+        kubectl annotate ingressclass nginx "ingressclass.kubernetes.io/is-default-class=true"
+        kubectl -n ingress-nginx wait --timeout=300s --for=condition=Available deployments --all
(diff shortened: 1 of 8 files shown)


.github/actions/kind/action.yml
```diff
@@ -47,3 +47,12 @@ runs:
         echo "KinD cluster:"
         kubectl cluster-info
         kubectl describe nodes
+
+    - name: Install Ingress controller
+      shell: bash
+      run: |
+        VERSION=controller-v1.6.4
```
anishasthana, line 54: Should we move this variable to something that's definied at the base level of the Makefile, and then pass it down to test/e2e/kind.sh as well?

anishasthana (commented):
Just had one comment, really.
Somewhat related, once https://github.com/project-codeflare/codeflare-sdk/issues/174 is merged in we should revisit the route related test cases

.github/actions/kind/action.yml
astefanutti: I think that's more an testing internal detail, than something that should be parameterised as for the core dependencies.

astefanutti:
> Just had one comment, really.
> Somewhat related, once [project-codeflare/codeflare-sdk#174](https://github.com/project-codeflare/codeflare-sdk/issues/174) is merged in we should revisit the route related test cases

Yes, that should be revised as the `TestMNISTRayClusterSDK` test is enabled on vanilla Kubernetes.

anishasthana (commented):
/lgtm

sutaakar (approved):
Good job

anishasthana:
Weird that the approved label wasn't applied.
/approve

PR #216, opened 2023-08-08
Operator Redesign

### LINKED ISSUE #281: Initialise and source configuration from a ConfigMap
## WHY

Part of the CodeFlare operator re-design, as described in https://github.com/project-codeflare/adr/pull/9.

Follows up #273.

## WHAT

The structured configuration should be sourced from a ConfigMap. If it doesn't exist already, the operator should initialise it with sensible defaults when it starts.

To be done on top of #216.

astefanutti:
Part of #279.

Closes #281.


config/manifests/bases/codeflare-operator.clusterserviceversion.yaml
```diff
@@ -14,16 +14,21 @@ spec:
...
+    - description: SchedulingSpec is the Schema for the SchedulingSpec API
+      displayName: SchedulingSpec
+      kind: SchedulingSpecs
+      name: schedulingspecs.workload.codeflare.dev
+      version: v1beta1
+    - description: QuotaSubtree is the Schema for the QuotaSubtree API
+      displayName: QuotaSubtree
+      kind: QuotaSubtrees
+      name: quotasubtrees.quota.codeflare.dev
+      version: v1
```
anishasthana, line 31: I feel like quotasubtrees should _not_ be a v1 version, but it's probably too late at this point for that.

config/rbac/role.yaml
```diff
@@ -18,18 +18,6 @@ rules:
   - patch
   - update
   - watch
-- apiGroups:
-  - apps
-  resources:
-  - deployments
```
anishasthana, line 24: I'm curious -- MCAD can still create deployments right? So won't we still need deployment perms for the CFO?

main.go
```diff
@@ -17,103 +17,209 @@ limitations under the License.
...
+		RetryPeriod:                &cfg.LeaderElection.RetryPeriod.Duration,
+		RenewDeadline:              &cfg.LeaderElection.RenewDeadline.Duration,
 	})
-	if err != nil {
-		setupLog.Error(err, "unable to start manager")
+	exitOnError(err, "unable to start manager")
+
+	mcadQueueController := mcad.NewJobController(mgr.GetConfig(), cfg.MCAD, &mcadconfig.MCADConfigurationExtended{})
+	if mcadQueueController == nil {
+		// FIXME
```
anishasthana, line 139: Can you elaborate on this?

pkg/config/config.go
```diff
@@ -0,0 +1,98 @@
...
+	// +optional
+	Health HealthConfiguration `json:"health,omitempty"`
+
+	// LeaderElection is the LeaderElection config to be used when configuring
+	// the manager.Manager leader election
+	LeaderElection *configv1alpha1.LeaderElectionConfiguration `json:"leaderElection,omitempty"`
+}
+
+type ClientConnection struct {
+	// QPS controls the number of queries per second allowed before client-side throttling
```
anishasthana, line 66: Neat. I've not thought of/seen this before

test/e2e/setup.sh
```diff
@@ -61,13 +51,10 @@ metadata:
   name: mcad-controller-rayclusters
 subjects:
   - kind: ServiceAccount
-    name: mcad-controller-mcad
-    namespace: codeflare-system
+    name: codeflare-operator-controller-manager
```
KPostOffice, line 54: I'm not seeing this service account anywhere else in the code base. Where is it coming from?

Makefile
```diff
@@ -222,15 +167,15 @@ vet: ## Run go vet against code.
 
 .PHONY: modules
 modules: ## Update Go dependencies.
-	go get $(MCAD_REPO)@$(MCAD_VERSION)
-	go get github.com/ray-project/kuberay/ray-operator
+	#go get $(MCAD_REPO)@$(MCAD_VERSION)
```
KPostOffice, line 170: Can we just remove this make directive?

KPostOffice (commented):
looks good to me, just a couple of questions

.github/workflows/verify_generated_files.yml
```diff
@@ -15,28 +15,6 @@ on:
       - '**go.mod'
       - '**go.sum'
 jobs:
-  verify-generated-functions:
```
sutaakar, line 18: What was the reason to delete this code?

sutaakar:
Looks good to me once comments are addressed

config/rbac/role.yaml
astefanutti: Right. The diff makes it hard to see, but this was actually duplicated, and there are still permissions for Deployments a bit below.

test/e2e/setup.sh
astefanutti: That ServiceAccount is defined here: https://github.com/project-codeflare/codeflare-operator/blob/9b2d77c378842a31b965ea4d2a5a5d18328c48c8/config/rbac/service_account.yaml#L4

But its name is changed to `codeflare-operator-controller-manager` by Kustomize global name prefixing.

.github/workflows/verify_generated_files.yml
astefanutti: It's deleted as the entire client generation has become obsolete and been removed.

main.go
astefanutti: I've updated the FIXME with more context.

Makefile
astefanutti: Good catch, thanks. It was a temporary work-around until all the pieces finally clicked. Updated in a243e589b40e1523cee6660ed83cc5f1a3690a91.

anishasthana (commented):
/lgtm from my perspective. We can merge it once @KPostOffice and @sutaakar  have approved it. (I'm not approving it myself as that would result in the PR merging)

KPostOffice:
this /lgtm as well

dimakis (commented):
LGTM 

great work Antonin 
the e2e tests fail for me locally but seen to pass in the CI so it may just be my env

anishasthana (approved):

PR #231, opened 2023-08-10
Update CodeFlare stack release instructions

anishasthana:


anishasthana:
@sutaakar  Raising this PR, I was thinking... Maybe we should update the release action to allow users to select _which_ components are going to have releases. This would lessen the burden on users (as opposed to forcing them to manually go to each repo and run the corresponding action)

sutaakar:
@anishasthana It can be done, though it will need some adjustments in workflows.
Just to be sure, the release strategy remains the same, right? By default release all components, occasionally for hotfix release just subset of components.

anishasthana:
>Just to be sure, the release strategy remains the same, right? By default release all components, occasionally for hotfix release just subset of components.

Yep! That's where my head is at. Something like a checkbox would probably suffice

KPostOffice (commented):
/lgtm

KPostOffice (approved):

PR #258, opened 2023-08-30
Fix Instascale Version

Fiona-Waters:
While deploying the codeflare stack on my osd cluster, the instascale pod did not come up and showed an `errImagePull` error. I noticed in the yaml that the image url was not as expected i.e it was v.0.0.7 instead of v0.0.7 as in Quay.io. This PR is updating the instascale image url.


anishasthana (approved):
/lgtm

Issue #51, opened 2023-03-06
Update MCAD Custom Resource to allow users to specific memory and cpu resources


anishasthana:

tedhtchang:
This issue is being worked on by [PR #63](https://github.com/project-codeflare/codeflare-operator/pull/63)

tedhtchang:
PR merged. So closing the issue.

PR #271, opened 2023-09-04
Adding instascale e2e test

Fiona-Waters:
DRAFT PR - Work In Progress
# Issue link
https://github.com/project-codeflare/instascale/issues/66

# What changes have been made

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


test/e2e/instascale_test.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+	config, err := test.Client().Core().CoreV1().ConfigMaps(namespace.Name).Create(test.Ctx(), config, metav1.CreateOptions{})
+	test.Expect(err).NotTo(HaveOccurred())
+	test.T().Logf("Created ConfigMap %s/%s successfully", config.Namespace, config.Name)
+
+	// create OCM connection
+	instascaleOCMSecret, err := test.Client().Core().CoreV1().Secrets("default").Get(test.Ctx(), "instascale-ocm-secret", metav1.GetOptions{})
+	if err != nil {
+		test.T().Errorf("unable to retrieve instascale-ocm-secret - Error : %v", err)
+	}
```
sutaakar, line 49: The test is expecting to be executed in OSD environment, right?

Fiona-Waters: Yes, I'm testing on OSD, ROSA and OCP which all use this secret.

Fiona-Waters: I believe OCP doesn't need the secret so will have to refactor to take that into account.

test/e2e/instascale_machinepool_test.go
```diff
@@ -0,0 +1,73 @@
...
+func TestInstascaleMachinePool(t *testing.T) {
+
+	test := With(t)
+	test.T().Parallel()
+
+	namespace := test.NewTestNamespace()
+
+	// Test configuration
+	config, err := TestConfig(test, namespace.Name)
+	test.Expect(err).To(BeNil())
```
sutaakar, line 23: It may be better to use `test.Expect(err).NotTo(HaveOccurred())` instead as is done in other places.

test/support/clusterpools.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+func CreateOCMConnection(secret string) (*ocmsdk.Connection, error) {
+	logger, err := ocmsdk.NewGoLoggerBuilder().
+		Debug(false).
+		Build()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Can't build logger: %v\n", err)
+		os.Exit(1)
+	}
```
sutaakar, line 35: What about returning `nil, err` here? To leave error processing to the upper layer.

test/support/clusterpools.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func MachinePoolsExist(connection *ocmsdk.Connection) (bool, error) {
+	machinePools := connection.ClustersMgmt().V1().Clusters().Cluster(ClusterID).MachinePools()
+	return machinePools != nil, nil
+}
+
+func NodePoolsExist(connection *ocmsdk.Connection) (bool, error) {
+	nodePools := connection.ClustersMgmt().V1().Clusters().Cluster(ClusterID).NodePools()
+	return nodePools != nil, nil
```
sutaakar, line 55: Same here.

test/support/clusterpools.go
```diff
@@ -0,0 +1,115 @@
...
+}
```
sutaakar, line 56: It seems like these functions are not used, can they be deleted?

test/e2e/instascale_machinepool_test.go
Fiona-Waters: Updated in 3 places. Thank you

test/support/clusterpools.go
Fiona-Waters: Updated

test/support/clusterpools.go
```diff
@@ -0,0 +1,115 @@
...
+		Build()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Can't build logger: %v\n", err)
+		os.Exit(1)
+	}
+	connection, err := ocmsdk.NewConnectionBuilder().
+		Logger(logger).
+		Tokens(string(secret)).
+		Build()
+	if err != nil || connection == nil {
```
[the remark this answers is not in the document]
Fiona-Waters: Updated

test/e2e/instascale_machinepool_test.go
Fiona-Waters: They are no longer needed, I have removed them.

test/support/clusterpools.go
Fiona-Waters: They are no longer needed, I have removed them.

test/e2e/instascale.go
```diff
@@ -0,0 +1,177 @@
...
+								{
+									Name:      "test",
+									MountPath: "/test",
+								},
+								{
+									Name:      "workdir",
+									MountPath: "/workdir",
+								},
+							},
+							WorkingDir: "workdir",
```
sutaakar, line 95:
```suggestion
							WorkingDir: "workdir",
```

test/e2e/instascale.go
```diff
@@ -0,0 +1,177 @@
...
+			Completions: Ptr(int32(1)),
+			Parallelism: Ptr(int32(1)),
+			Template: corev1.PodTemplateSpec{
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						{
+							Name:  "job",
+							Image: GetPyTorchImage(),
+							Env: []corev1.EnvVar{
+								corev1.EnvVar{Name: "PYTHONUSERBASE", Value: "/test2"},
```
sutaakar, line 81:
```suggestion
								{Name: "PYTHONUSERBASE", Value: "/test2"},
```

ChristianZaccaria (changes requested):

test/support/support.go
```diff
@@ -30,9 +30,10 @@ import (
 var (
 	ApplyOptions = metav1.ApplyOptions{FieldManager: "codeflare-test", Force: true}
 
-	TestTimeoutShort  = 1 * time.Minute
-	TestTimeoutMedium = 2 * time.Minute
-	TestTimeoutLong   = 5 * time.Minute
+	TestTimeoutShort           = 1 * time.Minute
+	TestTimeoutMedium          = 2 * time.Minute
+	TestTimeoutLong            = 5 * time.Minute
+	TestTimeoutGpuProvisioning = 30 * time.Minute
```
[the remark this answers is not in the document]
sutaakar: Honestly, I am not a fan of configurable timeouts in general.
Default timeout should accommodate for all the used environments if possible.

About timeout generic naming - the problem I see is that every action with timeout is unique. By naming the timeouts generically you lose coupling between specific timeout. As a result if you need to adjust the timeout then you need to first search the code to see what timeout is used, also by raising shared timeout you also affect other tests, which may result in unintended longer test execution before fail.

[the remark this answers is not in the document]
ChristianZaccaria: That's a great point, thanks Karel, it makes sense for unique cases to have their own timeout. I was thinking that since this new timeout is coincidently the max amount when running [e2e-tests locally](https://github.com/project-codeflare/codeflare-operator/blob/de7fc9dce224a92282bf33665465b3a15909b55a/Makefile#L364-L366) for example, it could be seen as just the max timeout and used generically. But, I understand what you mean as well.

test/support/ocm.go
```diff
@@ -0,0 +1,52 @@
...
+	connection, err := buildOCMConnection(ocmToken)
+	test.Expect(err).NotTo(gomega.HaveOccurred())
+	return connection
+}
+
+func buildOCMConnection(secret string) (*ocmsdk.Connection, error) {
+	connection, err := ocmsdk.NewConnectionBuilder().
+		Tokens(secret).
+		Build()
+	if err != nil || connection == nil {
```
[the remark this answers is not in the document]
Fiona-Waters: This is probably left over from debugging, and I agree that it is unnecessary. I will remove it. Thanks Christian.

test/support/support.go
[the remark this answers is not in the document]
Fiona-Waters: Good call, can update.

test/support/codeflare.go
```diff
@@ -50,6 +59,24 @@ func GetPyTorchImage() string {
...
+
+func GetInstaScaleOcmSecretNamespace() string {
+	return lookupEnvOrDefault(InstaScaleOcmSecretNamespace, "default")
+}
+
+func GetOsdClusterId() (string, bool) {
+	return os.LookupEnv(OsdClusterID)
+}
+
+func IS_OSD() bool {
```
sutaakar, line 74:
```suggestion
func IsOsd() bool {
```

test/support/codeflare.go
```diff
@@ -50,6 +60,28 @@ func GetPyTorchImage() string {
...
+
+func GetOsdClusterId() (string, bool) {
+	return os.LookupEnv(OsdClusterID)
+}
+
+func IsOsd() bool {
+	env := lookupEnvOrDefault(IsOSD, "false")
+	osd, err := strconv.ParseBool(env)
+	if err != nil {
+		fmt.Errorf("error parsing IS_OSD variable, error: %v ", err)
```
ChristianZaccaria, line 79:
```suggestion
		fmt.Errorf("error parsing IsOSD variable, error: %v ", err)
```

sutaakar, line 79: Maybe would be better to use `test.Expect(err).NotTo(gomega.HaveOccurred(), fmt.Sprintf("error parsing IS_OSD variable, error: %v ", err))` here. That will make sure the error is propagated to the output.

ChristianZaccaria (changes requested):

test/support/codeflare.go
sutaakar, line 79: May be better to use `test.T().Logf("error parsing IS_OSD variable, using 'false' value, error: %v ", err)`

sutaakar: That way the error gets printed to the output

sutaakar: to be precise - `...IS_OSD environment variable...`

Fiona-Waters: Updating this somewhat but leaving it as IS_OSD as it is referring to the environment variable.

sutaakar:
@astefanutti @dimakis Can you please take a look?

test/support/codeflare.go
```diff
@@ -50,6 +61,30 @@ func GetPyTorchImage() string {
 	return lookupEnvOrDefault(CodeFlareTestPyTorchImage, "pytorch/pytorch:1.11.0-cuda11.3-cudnn8-runtime")
 }
 
+func GetInstaScaleOcmSecretName() string {
+	return lookupEnvOrDefault(InstaScaleOcmSecretName, "instascale-ocm-secret")
+}
+
+func GetInstaScaleOcmSecretNamespace() string {
+	return lookupEnvOrDefault(InstaScaleOcmSecretNamespace, "default")
+}
```
astefanutti, line 70: Can this be collapsed into a single environment variable, so it's referenced as `default/instascale-ocm-secret` for example.

test/support/codeflare.go
```diff
@@ -30,8 +33,16 @@ const (
...
 	CodeFlareTestOutputDir = "CODEFLARE_TEST_OUTPUT_DIR"
+
+	// The name of a secret containing InstaScale OCM token.
+	InstaScaleOcmSecretName = "INSTASCALE_OCM_SECRET_NAME"
+	// The namespace where a secret containing InstaScale OCM token is stored.
+	InstaScaleOcmSecretNamespace = "INSTASCALE_OCM_SECRET_NAMESPACE"
+	// Cluster ID for OSD cluster used in tests, used for testing InstaScale
+	OsdClusterID = "CLUSTERID"
+	// Determine if test is being run on an OSD cluster, used for testing InstaScale.
+	IsOSD = "IS_OSD"
```
astefanutti, line 45: Could it be assumed that it's an OSD cluster if an OSD cluster ID has been provided?

test/support/codeflare.go
```diff
@@ -30,8 +33,16 @@ const (
 	CodeFlareTestPyTorchImage = "CODEFLARE_TEST_PYTORCH_IMAGE"
 
 	// The testing output directory, to write output files into.
-
 	CodeFlareTestOutputDir = "CODEFLARE_TEST_OUTPUT_DIR"
+
```
astefanutti, line 37: I realise that file should probably be called `environment.go`, WDYT?

test/support/config_map.go
```diff
@@ -0,0 +1,45 @@
...
+package support
+
+import (
+	"github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func CreateConfigMap(t Test, namespace string, content map[string][]byte) *corev1.ConfigMap {
```
astefanutti, line 26: I'd suggest we move that function into the `core.go` file, as the idea was to organise the support files by API group.

test/e2e/instascale_machinepool_test.go
```diff
@@ -0,0 +1,88 @@
...
+	mcadv1beta1 "github.com/project-codeflare/multi-cluster-app-dispatcher/pkg/apis/controller/v1beta1"
+
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+
+	. "github.com/project-codeflare/codeflare-operator/test/support"
+)
+
+func TestInstascaleMachinePool(t *testing.T) {
+
```
astefanutti, line 32: nit: extra return can be removed.

test/e2e/instascale_machinepool_test.go
```diff
@@ -0,0 +1,88 @@
...
+		// pip requirements
+		"requirements.txt": ReadFile(test, "mnist_pip_requirements.txt"),
+		// MNIST training script
+		"mnist.py": ReadFile(test, "mnist.py"),
+	}
+	cm := CreateConfigMap(test, namespace.Name, testConfigData)
+
+	//create OCM connection
+	connection := CreateOCMConnection(test)
+
```
astefanutti, line 53: nit: extra return can be removed.

test/e2e/instascale_machinepool_test.go
```diff
@@ -0,0 +1,88 @@
...
+	test.T().Parallel()
+
+	if !IsOsd(test) {
+		test.T().Skip("Skipping test as not running on an OSD cluster")
+	}
+
+	namespace := test.NewTestNamespace()
+
+	// Test configuration
+	testConfigData := map[string][]byte{
```
astefanutti, line 43: nit: Maybe that can be inlined and the `testConfigData` variable removed.

test/e2e/instascale_app_wrapper.go
```diff
@@ -0,0 +1,148 @@
...
+				},
+			},
+		},
+	}
+
+	_, err := test.Client().MCAD().WorkloadV1beta1().AppWrappers(namespace.Name).Create(test.Ctx(), aw, metav1.CreateOptions{})
+	test.Expect(err).NotTo(HaveOccurred())
+	test.T().Logf("AppWrapper created successfully %s/%s", aw.Namespace, aw.Name)
+
+	test.Eventually(AppWrapper(test, namespace, aw.Name), TestTimeoutGpuProvisioning).
```
astefanutti, line 144: I'd be inclined to move that assertion in the test case itself, so that support function is strictly scoped to creating an AppWrapper.

test/e2e/instascale_machinepool_test.go
```diff
@@ -0,0 +1,88 @@
...
+	test.T().Logf("Waiting for Job %s/%s to complete", job.Namespace, job.Name)
+	test.Eventually(Job(test, job.Namespace, job.Name), TestTimeoutLong).Should(
+		Or(
+			WithTransform(ConditionStatus(batchv1.JobComplete), Equal(corev1.ConditionTrue)),
+			WithTransform(ConditionStatus(batchv1.JobFailed), Equal(corev1.ConditionTrue)),
+		))
+
+	// Assert the job has completed successfully
+	test.Expect(GetJob(test, job.Namespace, job.Name)).
+		To(WithTransform(ConditionStatus(batchv1.JobComplete), Equal(corev1.ConditionTrue)))
```
astefanutti, line 79: Out of curiosity, what's the added value of asserting the underlying Job? Could asserting the completion of the AppWrapper be enough?

test/e2e/instascale_app_wrapper.go
```diff
@@ -0,0 +1,148 @@
...
+			Template: corev1.PodTemplateSpec{
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						{
+							Name:  "job",
+							Image: GetPyTorchImage(),
+							Env: []corev1.EnvVar{
+								{Name: "PYTHONUSERBASE", Value: "/workdir"},
+							},
+							Command: []string{"/bin/sh", "-c", "pip install -r /test/requirements.txt && torchrun /test/mnist.py"},
```
astefanutti, line 54: Just making sure @Srihari1192 is aware of it in the context of #51

astefanutti (commented):
Very well done! Just left of couple of comments, otherwise looks very good!

test/support/codeflare.go
Fiona-Waters: Maybe but would it make sense when the variables are used separately in the Secret(namespace) and Get(secret name) parts here: 
```
instascaleOCMSecret, err := test.Client().Core().CoreV1().Secrets(GetInstaScaleOcmSecretNamespace()).Get(test.Ctx(), GetInstaScaleOcmSecretName(), metav1.GetOptions{})
```

test/e2e/instascale_machinepool_test.go
Fiona-Waters: You're right it's not necessary, I will remove it.

test/support/codeflare.go
Fiona-Waters: Yes I think so, I have updated to reflect this.

test/support/codeflare.go
Fiona-Waters: Updated

test/support/config_map.go
Fiona-Waters: Updated

test/e2e/instascale_machinepool_test.go
Fiona-Waters: Updated

test/e2e/instascale_machinepool_test.go
Fiona-Waters: Updated

test/e2e/instascale_machinepool_test.go
Fiona-Waters: Updated

test/support/clusterpools.go
```diff
@@ -0,0 +1,44 @@
...
+package support
+
+import (
+	"github.com/onsi/gomega"
+
+	ocmsdk "github.com/openshift-online/ocm-sdk-go"
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+)
+
+func MachinePools(t Test, connection *ocmsdk.Connection) func(g gomega.Gomega) []*cmv1.MachinePool {
```
astefanutti, line 26: Could the content of that file be moved to the `ocm.go` file? It seems it's all OCM related importing from `ocm-sdk-go`?

Fiona-Waters: Yes that makes sense. Moved.

test/support/codeflare.go
astefanutti: I'd rather have a single env variable for clients to set, e.g. `INSTASCALE_OCM_SECRET=default/instacale-ocm-secret` and the two functions adapted so they split the provided string, or a single function that returns both strings, e.g.:

```go
func GetInstaScaleOcmSecret() (string, string) {
	res := strings.SplitN(lookupEnvOrDefault(InstaScaleOcmSecret, "default/instascale-ocm-secret"), "/", 2)
	return res[0], res[1]
}
```

Fiona-Waters: Ok, thank you. I have updated.

astefanutti:
/lgtm

Great work!

PR #276, opened 2023-09-06
Use sed command to replace MCAD repo reference in Makefile

sutaakar:
# Issue link

# What changes have been made

Kustomize fn run is using Docker. This approach is incompatible with OpenShift CI, which doesn't provide Docker environment.
To make operator deployment being OpenShift CI friendly (and to remove a need to have specifically Docker installed for various Makefile actions) replacing the function with sed command providing same functionality.

# Verification steps

Execute for example `make install` without Docker.

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -249,26 +249,26 @@ endif
 
 .PHONY: install
 install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) fn run config/crd/mcad --image gcr.io/kpt-fn/apply-setters:v0.2.0 -- MCAD_CRD=$(MCAD_CRD)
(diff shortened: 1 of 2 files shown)


sutaakar:
@astefanutti 
`config/crd/mcad/kustomization.yaml` contains placeholder, I have left there kustomize metadata as it doesn't block or affect anything
The value itself is replaced just when needed (deployment using Makefile or bundle build).

sutaakar:
That would be the best :)

astefanutti:
> That would be the best :)

Maybe a quick note in the README can be added, WDYT?

sutaakar:
@astefanutti added a note in readme

PR #277, opened 2023-09-06
Remove irrelevant external metrics APIService resources

### LINKED ISSUE #247: Deploy MCAD in agent mode (without multi-cluster support)
## WHY

Multi-cluster mode provided by MCAD is experimental and is not used, aside for the purpose of its development.

Even if multi-cluster mode is deactivated by default, a number of resources are installed only for the purpose of its operation.

These resources are used to serve external metrics by the embedded external metrics server, that's not started in single cluster mode, i.e., APIService resources, and a number of ClusterRoleBinding resources, that are irrelevant to the single cluster setup, and introduce noise for security scanning.

## WHAT

The CodeFlare operator should not install those resources, as the multi-cluster mode is still experimental and not supported.

## HOW

The corresponding resources should be removed from the installation.

## TESTS

The e2e tests should pass.

ChristianZaccaria:
# Issue link

Closes #247 

# What changes have been made

As mentioned in the issue: The multi-cluster mode provided by MCAD is experimental and is not used. Even if the multi-cluster mode is deactivated by default, a number of resources are installed only for the purpose of its operation.

For this reason, we are removing the APIService resources and ClusterRoleBindings that are irrelevant to the single cluster setup.

The removal of the CRBs was done as part of this [PR #270 ](https://github.com/project-codeflare/codeflare-operator/pull/270)

## Checks
- [ ]  e2e tests should continue to pass.


controllers/multi_cluster_app_dispatcher.go
```diff
@@ -22,7 +22,6 @@ import (
 
 var multiClusterAppDispatcherTemplates = []string{
 	"mcad/configmap.yaml.tmpl",
-	"mcad/service.yaml.tmpl",
 	"mcad/serviceaccount.yaml.tmpl",
 	"mcad/deployment.yaml.tmpl",
```
astefanutti, line 26: I think the corresponding ports from the Deployment can also be removed:

```yaml
ports:
  - name: https
     containerPort: 6443
     protocol: TCP
  - name: http
     containerPort: 8080
    protocol: TCP
```

ChristianZaccaria: Great catch! Changes were made now. 

I do have a few thoughts though. Should we also be removing items from the [MCAD test data](https://github.com/project-codeflare/codeflare-operator/tree/5209f7a7cea17b303127bd5132ec9e02a476ab84/controllers/testdata) ?  

I.e., in mcad_test_results, there are some rolebindings and the same mentioned ports in the deployment of these tests.

These tests seem to run on the `tag-and-build` workflow.

astefanutti: Great catch too! Indeed, these test data files should be amended accordingly too.

ChristianZaccaria: Excellent, the respective files have been removed. Waiting on tests to pass.

Issue #275, opened 2023-09-06
Implement exposeService method in E2E tests support package


jiripetrlik:
## WHY
New util method will contain logic which will help us to create routes and expose services using these routes.
## WHAT
New exposeService function will:
- accept namespace
- accept service name
- create route
- wait for route to be ready

PR #284, opened 2023-09-12
[275] Implement ExposeService method

jiripetrlik:
# Issue link
Issue: https://github.com/project-codeflare/codeflare-operator/issues/275

PR to distributed-workloads with changes to tests will follow.

### DIFF
--- a/test/support/ingress.go
+++ b/test/support/ingress.go
@@ -17,6 +17,8 @@ limitations under the License.
 package support
 
 import (
+	"net/url"
+
 	"github.com/onsi/gomega"
 
 	networkingv1 "k8s.io/api/networking/v1"
@@ -38,4 +40,62 @@ func GetIngress(t Test, namespace, name string) *networkingv1.Ingress {
 
 func LoadBalancerIngresses(ingress *networkingv1.Ingress) []networkingv1.IngressLoadBalancerIngress {
 	return ingress.Status.LoadBalancer.Ingress
+}
+
+func ExposeServiceByIngress(t Test, name string, namespace string, serviceName string, servicePort string) url.URL {
+	ingress := &networkingv1.Ingress{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: networkingv1.SchemeGroupVersion.String(),
+			Kind:       "Ingress",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Annotations: map[string]string{
+				"nginx.ingress.kubernetes.io/use-regex":      "true",
+				"nginx.ingress.kubernetes.io/rewrite-target": "/$2",
+			},
+		},
+		Spec: networkingv1.IngressSpec{
+			Rules: []networkingv1.IngressRule{
+				networkingv1.IngressRule{
+					IngressRuleValue: networkingv1.IngressRuleValue{
+						HTTP: &networkingv1.HTTPIngressRuleValue{
+							Paths: []networkingv1.HTTPIngressPath{
+								networkingv1.HTTPIngressPath{
+									Path:     "/" + name + "(/|$)(.*)",
+									PathType: Ptr(networkingv1.PathTypePrefix),
+									Backend: networkingv1.IngressBackend{
+										Service: &networkingv1.IngressServiceBackend{
+											Name: serviceName,
+											Port: networkingv1.ServiceBackendPort{
+												Name: servicePort,
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	_, err := t.Client().Core().NetworkingV1().Ingresses(ingress.Namespace).Create(t.Ctx(), ingress, metav1.CreateOptions{})
+	t.Expect(err).NotTo(gomega.HaveOccurred())
+	t.T().Logf("Created Ingress %s/%s successfully", ingress.Namespace, ingress.Name)
+
+	t.T().Logf("Waiting for Ingress %s/%s to be admitted", ingress.Namespace, ingress.Name)
+	t.Eventually(Ingress(t, ingress.Namespace, ingress.Name), TestTimeoutMedium).
+		Should(gomega.WithTransform(LoadBalancerIngresses, gomega.HaveLen(1)))
+
+	ingress = GetIngress(t, ingress.Namespace, ingress.Name)
+
+	ingressURL := url.URL{
+		Scheme: "http",
+		Host:   ingress.Status.LoadBalancer.Ingress[0].IP,
+		Path:   name,
+	}
+	return ingressURL
 }
--- a/test/support/route.go
+++ b/test/support/route.go
@@ -17,9 +17,14 @@ limitations under the License.
 package support
 
 import (
+	"net/http"
+	"net/url"
+
 	"github.com/onsi/gomega"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 
 	routev1 "github.com/openshift/api/route/v1"
 )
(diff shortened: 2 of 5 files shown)


test/support/route.go
```diff
@@ -36,3 +40,45 @@ func GetRoute(t Test, namespace, name string) *routev1.Route {
...
+	// Retrieve hostname
+	r, err = t.Client().Route().RouteV1().Routes(r.Namespace).Get(t.Ctx(), r.Name, metav1.GetOptions{})
+	t.Expect(err).NotTo(gomega.HaveOccurred())
+	hostname := r.Status.Ingress[0].Host
+
+	// Wait for expected HTTP code
+	t.Eventually(func() int {
+		resp, _ := http.Get("http://" + hostname)
+		return resp.StatusCode
+	}, TestTimeoutLong).Should(gomega.Equal(expetedCode))
```
sutaakar, line 81: What about waiting until Route stops returning HTTP 503 and a text in response body referring to the Pods being down?
That would allow dropping expected code parameter.

test/support/route.go
```diff
@@ -36,3 +40,45 @@ func GetRoute(t Test, namespace, name string) *routev1.Route {
 	t.T().Helper()
 	return Route(t, namespace, name)(t)
 }
+
+func ExposeService(t Test, namespace string, service string, route string, port string, expetedCode int) *routev1.Route {
```
sutaakar, line 44: The parameter names may be misleading, what about:
serviceName
servicePort
routeName

jiripetrlik:
@sutaakar 
Thank you for the review. I've tried to address both your comments.

test/support/route.go
```diff
@@ -36,3 +40,45 @@ func GetRoute(t Test, namespace, name string) *routev1.Route {
 	t.T().Helper()
 	return Route(t, namespace, name)(t)
 }
+
+func ExposeService(t Test, namespace string, serviceName string, routeName string, servicePort string) *routev1.Route {
```
sutaakar, line 44: A nitpick: 
```suggestion
func ExposeService(t Test, namespace string, serviceName, servicePort, routeName string) *routev1.Route {
```
To keep service related properties together

test/support/route.go
```diff
@@ -36,3 +40,45 @@ func GetRoute(t Test, namespace, name string) *routev1.Route {
...
+	t.Eventually(Route(t, r.Namespace, r.Name), TestTimeoutLong).
+		Should(gomega.WithTransform(ConditionStatus(routev1.RouteAdmitted), gomega.Equal(corev1.ConditionTrue)))
+
+	// Retrieve hostname
+	r, err = t.Client().Route().RouteV1().Routes(r.Namespace).Get(t.Ctx(), r.Name, metav1.GetOptions{})
+	t.Expect(err).NotTo(gomega.HaveOccurred())
+	hostname := r.Status.Ingress[0].Host
+
+	// Wait for expected HTTP code
+	t.Eventually(func() int {
```
sutaakar, line 78: Can you also check that response body contains text specific to the Route not serving yet, for example:
`Route and path matches, but all pods are down.`

test/support/route.go
sutaakar: Also makes me think, do we need to specify route name? Is there any case when caller would need this info?
What about using service name as route name, possibly with some suffix? That way we reasonably assure that Route name is unique.

test/support/route.go
sutaakar: The reason is that theoretically the 503 can be a correct response code for the service.
Though that may be really edge case.

test/support/ingress.go
```diff
@@ -39,3 +39,43 @@ func GetIngress(t Test, namespace, name string) *networkingv1.Ingress {
 func LoadBalancerIngresses(ingress *networkingv1.Ingress) []networkingv1.IngressLoadBalancerIngress {
 	return ingress.Status.LoadBalancer.Ingress
 }
+
+func exposeServiceIngress(t Test, ingressName string, namespace string, serviceName string, servicePort string) *networkingv1.Ingress {
+	prefixType := networkingv1.PathTypePrefix
+	ingress := &networkingv1.Ingress{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: networkingv1.SchemeGroupVersion.String(),
+			Kind:       "Route",
```
sutaakar, line 48:
```suggestion
			Kind:       "Ingress",
```

test/support/test_parameters.go
```diff
@@ -0,0 +1,34 @@
...
+package support
+
+import (
+	"os"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
```
sutaakar, line 26: My idea was to have here just constants, not variables.
Something like https://github.com/opendatahub-io/distributed-workloads/blob/8b189dc847fe56f7bf57d69c002a5d85defe3c40/tests/integration/config.go

test/support/service.go
```diff
@@ -0,0 +1,27 @@
...
+
+package support
+
+func ExposeService(t Test, name string, namespace string, serviceName string, servicePort string) string {
+	if IsOpenShift(t) {
+		route := ExposeServiceRoute(t, namespace, serviceName, name, servicePort)
+		return "http://" + route.Status.Ingress[0].Host
+	} else {
+		exposeServiceIngress(t, namespace, serviceName, name, servicePort)
+		return KubernetesHostname + "/" + name
```
sutaakar, line 25:
```suggestion
		return "http://" + KubernetesHostname + "/" + name
```

test/support/route.go
sutaakar: Currently not needed.

test/support/ingress.go
```diff
@@ -39,3 +39,43 @@ func GetIngress(t Test, namespace, name string) *networkingv1.Ingress {
...
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
```
sutaakar, line 79: Can you create the Ingress on cluster and make sure it is ready? Same like it is done in Route.

sutaakar:
Please use the new function in https://github.com/project-codeflare/codeflare-operator/blob/a14914da37d61dd2e3de160c7033e3b2fe6d500e/test/e2e/mnist_rayjob_mcad_raycluster_test.go to make sure it works properly.
Also the `KUBERNETES_HOSTNAME` needs to be set in PR check workflow.

jiripetrlik:
Hello @sutaakar , I've tried to address all the comments. In the final version I removed KUBERNETES_HOSTNAME parameter at all.

test/support/service.go
```diff
@@ -0,0 +1,41 @@
...
+*/
+
+package support
+
+import "net/url"
+
+func ExposeService(t Test, name string, namespace string, serviceName string, servicePort string) url.URL {
+	if IsOpenShift(t) {
+		route := ExposeServiceRoute(t, name, namespace, serviceName, servicePort)
+		route = GetRoute(t, route.Namespace, route.Name)
```
sutaakar, line 24: This can be moved to `ExposeServiceRoute`, same as done in Ingress.

test/support/ingress.go
```diff
@@ -39,3 +39,56 @@ func GetIngress(t Test, namespace, name string) *networkingv1.Ingress {
 func LoadBalancerIngresses(ingress *networkingv1.Ingress) []networkingv1.IngressLoadBalancerIngress {
 	return ingress.Status.LoadBalancer.Ingress
 }
+
+func ExposeServiceIngress(t Test, name string, namespace string, serviceName string, servicePort string) *networkingv1.Ingress {
```
sutaakar, line 43: Optional suggestion - what about using name `ExposeServiceByIngress`

test/support/route.go
```diff
@@ -36,3 +40,45 @@ func GetRoute(t Test, namespace, name string) *routev1.Route {
 	t.T().Helper()
 	return Route(t, namespace, name)(t)
 }
+
+func ExposeServiceRoute(t Test, name string, namespace string, serviceName string, servicePort string) *routev1.Route {
```
sutaakar, line 44: Same here.

test/support/service.go
```diff
@@ -0,0 +1,41 @@
...
+
+		serviceURL := url.URL{
+			Scheme: "http",
```
sutaakar, line 27: A nitpick, can you derive using HTTP or HTTPS based on Route spec or status content?

test/support/route.go
```diff
@@ -36,3 +41,57 @@ func GetRoute(t Test, namespace, name string) *routev1.Route {
...
+		resp, _ := http.Get("http://" + hostname)
+		return resp.StatusCode
+	}, TestTimeoutLong).Should(gomega.Not(gomega.Equal(503)))
+
+	r = GetRoute(t, r.Namespace, r.Name)
+
+	scheme := "http"
+	if r.Spec.Port.TargetPort.String() == "https" {
+		scheme = "https"
+	}
```
sutaakar, line 89: When I think about it again, this is not actually needed, as above we construct HTTP route anyway, so we are sure it is HTTP.
Can you please remove this code?

sutaakar (approved):
/lgtm

PR #285, opened 2023-09-13
e2e support: return AppWrapper pointers instead of structs

sutaakar:
# Issue link

Intention is to align `AppWrappers` function with `AppWrapper`, so result of both functions can be used in transforming functions.

# What changes have been made

`AppWrappers` function now returns slice of pointers on `AppWrapper` instead of slice of structs.

# Verification steps

Tested in ODH DW tests.

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/test/support/mcad.go
+++ b/test/support/mcad.go
@@ -32,11 +32,18 @@ func AppWrapper(t Test, namespace *corev1.Namespace, name string) func(g gomega.
 	}
 }
 
-func AppWrappers(t Test, namespace *corev1.Namespace) func(g gomega.Gomega) []mcadv1beta1.AppWrapper {
-	return func(g gomega.Gomega) []mcadv1beta1.AppWrapper {
+func AppWrappers(t Test, namespace *corev1.Namespace) func(g gomega.Gomega) []*mcadv1beta1.AppWrapper {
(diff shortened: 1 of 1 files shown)


test/support/mcad.go
```diff
@@ -32,11 +32,17 @@ func AppWrapper(t Test, namespace *corev1.Namespace, name string) func(g gomega.
...
-	return func(g gomega.Gomega) []mcadv1beta1.AppWrapper {
+func AppWrappers(t Test, namespace *corev1.Namespace) func(g gomega.Gomega) []*mcadv1beta1.AppWrapper {
+	return func(g gomega.Gomega) []*mcadv1beta1.AppWrapper {
 		aws, err := t.Client().MCAD().WorkloadV1beta1().AppWrappers(namespace.Name).List(t.Ctx(), metav1.ListOptions{})
 		g.Expect(err).NotTo(gomega.HaveOccurred())
-		return aws.Items
+
+		awsp := []*mcadv1beta1.AppWrapper{}
+		for _, v := range aws.Items {
+			awsp = append(awsp, &v)
```
astefanutti, line 42: This should be:

```go
for i, v := range aws.Items {
	v := v
	awsp = append(awsp, &v)
```

See https://github.com/golang/go/discussions/56010

Issue #184, opened 2023-07-18
Create upgrade e2e scenario covering upgrade during training


sutaakar:
Scenario should start a training, perform an operator upgrade during training and finish the training.

Covered features:
- ability of the new operator to process CRs created by old operator
- ability of CodeFlare SDK to connect to existing Cluster and work with AppWrapper created by previous version  (i.e. one client session creates an AppWrapper, a later client session can delete/get the AppWrapper)

sutaakar:
@astefanutti What do you think about this scenario?

astefanutti:
@sutaakar I agree it'd be critical to assess existing operands are not affected by the upgrade, especially those that are not in terminal states.

PR #286, opened 2023-09-14
Create OLM upgrade e2e scenario using codeflare SDK

Srihari1192:
# Issue link

https://github.com/project-codeflare/codeflare-operator/issues/184
# What changes have been made

- Added TestMNISTRayClusterUp and TestMnistJobSubmit in OlM upgrade test to run the test before an operator upgrade and after upgrade 
- Added methods CreateTestNamespaceWithName and DeleteTestNamespace in namespace support class for OLM upgrade tests

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests


test/e2e/mnist_rayjob.py
```diff
@@ -0,0 +1,44 @@
...
+from time import sleep
+
+from torchx.specs.api import AppState, is_terminal
+
+from codeflare_sdk.cluster.cluster import Cluster, ClusterConfiguration
+from codeflare_sdk.job.jobs import DDPJobDefinition
+
+namespace = sys.argv[1]
+
+cluster = Cluster(ClusterConfiguration(name='mnist'))
```
sutaakar, line 12:
```suggestion
cluster = Cluster(ClusterConfiguration(name='mnist', namespace=namespace))
```

test/e2e/olm_upgrade_test.go
```diff
@@ -0,0 +1,333 @@
...
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: corev1.SchemeGroupVersion.String(),
+				Kind:       "ServiceAccount",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "ray-user",
+				Namespace: namespaceName,
+			},
+		}
+		serviceAccount, err = test.Client().Core().CoreV1().ServiceAccounts(namespaceName).Create(test.Ctx(), serviceAccount, metav1.CreateOptions{})
```
sutaakar, line 260: This user is missing role to get Routes:
```

HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"routes.route.openshift.io is forbidden: User \"system:serviceaccount:test-ns-olmupgrade:ray-user\" cannot list resource \"routes\" in API group \"route.openshift.io\" in the namespace \"test-ns-olmupgrade\"","reason":"Forbidden","details":{"group":"route.openshift.io","kind":"routes"},"code":403}
```
You may reuse "sdk-user" from above for it.

test/e2e/olm_upgrade_test.go
```diff
@@ -0,0 +1,333 @@
...
+
+var namespaceName string = "test-ns-olmupgrade"
+var namespace *corev1.Namespace
+
+// Creates a Ray cluster
+func TestMNISTRayClusterUp(t *testing.T) {
+
+	test := With(t)
+	test.T().Parallel()
+	if os.Getenv("RUN_OLM_TESTS") == "true" {
```
sutaakar, line 43: Maybe it is better to have inverted condition here:
if os.Getenv("RUN_OLM_TESTS") != "true" {
	test.T().Skip("Skipping OLM upgarde test because RUN_OLM_TESTS is not set")
}

That way you can avoid having whole test as a condition branch.

test/e2e/olm_upgrade_test.go
Srihari1192: Done

test/e2e/olm_upgrade_test.go
Srihari1192: Modified to the inverted condition

Srihari1192:
@sutaakar OLM tests are Failing due to lack of resources in KinD cluster and test are pass in local.. I think we can enable this tests only when Large runners are available

sutaakar:
@Srihari1192 Last part of the log:
```
[notice] A new release of pip available: 22.3 -> 23.2.1
[notice] To update, run: pip install --upgrade pip
Written to: mnist.yaml
╭──────────────────────╮
│   🚀 Cluster Queue   │
│      Status 🚀       │
│ +-------+----------+ │
│ | Name  | Status   | │
│ +=======+==========+ │
│ | mnist | queueing | │
│ |       |          | │
│ +-------+----------+ │
╰──────────────────────╯
Waiting for requested resources to be set up...
No instances found, nothing to be done.
Traceback (most recent call last):
  File "raycluster_sdk.py", line 28, in <module>
    cluster.wait_ready()
  File "/opt/app-root/lib64/python3.8/site-packages/codeflare_sdk/cluster/cluster.py", line 273, in wait_ready
    dashboard_ready = self.is_dashboard_ready()
  File "/opt/app-root/lib64/python3.8/site-packages/codeflare_sdk/cluster/cluster.py", line 255, in is_dashboard_ready
    response = requests.get(self.cluster_dashboard_uri(), timeout=5)
  File "/opt/app-root/lib64/python3.8/site-packages/requests/api.py", line 73, in get
    return request("get", url, params=params, **kwargs)
  File "/opt/app-root/lib64/python3.8/site-packages/requests/api.py", line 59, in request
    return session.request(method=method, url=url, **kwargs)
  File "/opt/app-root/lib64/python3.8/site-packages/requests/sessions.py", line 573, in request
    prep = self.prepare_request(req)
  File "/opt/app-root/lib64/python3.8/site-packages/requests/sessions.py", line 484, in prepare_request
    p.prepare(
  File "/opt/app-root/lib64/python3.8/site-packages/requests/models.py", line 368, in prepare
    self.prepare_url(url, params)
  File "/opt/app-root/lib64/python3.8/site-packages/requests/models.py", line 439, in prepare_url
    raise MissingSchema(
requests.exceptions.MissingSchema: Invalid URL 'None': No scheme supplied. Perhaps you meant http://None?
```
Can you confirm whether SDK supports using Ingress? If so, can you check that Ingress is properly created when using KinD with Ingress installed in the test setup?

Srihari1192:
> @Srihari1192 Last part of the log:
> 
> ```
> [notice] A new release of pip available: 22.3 -> 23.2.1
> [notice] To update, run: pip install --upgrade pip
> Written to: mnist.yaml
> ╭──────────────────────╮
> │   🚀 Cluster Queue   │
> │      Status 🚀       │
> │ +-------+----------+ │
> │ | Name  | Status   | │
> │ +=======+==========+ │
> │ | mnist | queueing | │
> │ |       |          | │
> │ +-------+----------+ │
> ╰──────────────────────╯
> Waiting for requested resources to be set up...
> No instances found, nothing to be done.
> Traceback (most recent call last):
>   File "raycluster_sdk.py", line 28, in <module>
>     cluster.wait_ready()
>   File "/opt/app-root/lib64/python3.8/site-packages/codeflare_sdk/cluster/cluster.py", line 273, in wait_ready
>     dashboard_ready = self.is_dashboard_ready()
>   File "/opt/app-root/lib64/python3.8/site-packages/codeflare_sdk/cluster/cluster.py", line 255, in is_dashboard_ready
>     response = requests.get(self.cluster_dashboard_uri(), timeout=5)
>   File "/opt/app-root/lib64/python3.8/site-packages/requests/api.py", line 73, in get
>     return request("get", url, params=params, **kwargs)
>   File "/opt/app-root/lib64/python3.8/site-packages/requests/api.py", line 59, in request
>     return session.request(method=method, url=url, **kwargs)
>   File "/opt/app-root/lib64/python3.8/site-packages/requests/sessions.py", line 573, in request
>     prep = self.prepare_request(req)
>   File "/opt/app-root/lib64/python3.8/site-packages/requests/sessions.py", line 484, in prepare_request
>     p.prepare(
>   File "/opt/app-root/lib64/python3.8/site-packages/requests/models.py", line 368, in prepare
>     self.prepare_url(url, params)
>   File "/opt/app-root/lib64/python3.8/site-packages/requests/models.py", line 439, in prepare_url
>     raise MissingSchema(
> requests.exceptions.MissingSchema: Invalid URL 'None': No scheme supplied. Perhaps you meant http://None?
> ```
> 
> Can you confirm whether SDK supports using Ingress? If so, can you check that Ingress is properly created when using KinD with Ingress installed in the test setup?

Sure

Srihari1192:
> @Srihari1192 Last part of the log:
> 
> ```
> [notice] A new release of pip available: 22.3 -> 23.2.1
> [notice] To update, run: pip install --upgrade pip
> Written to: mnist.yaml
> ╭──────────────────────╮
> │   🚀 Cluster Queue   │
> │      Status 🚀       │
> │ +-------+----------+ │
> │ | Name  | Status   | │
> │ +=======+==========+ │
> │ | mnist | queueing | │
> │ |       |          | │
> │ +-------+----------+ │
> ╰──────────────────────╯
> Waiting for requested resources to be set up...
> No instances found, nothing to be done.
> Traceback (most recent call last):
>   File "raycluster_sdk.py", line 28, in <module>
>     cluster.wait_ready()
>   File "/opt/app-root/lib64/python3.8/site-packages/codeflare_sdk/cluster/cluster.py", line 273, in wait_ready
>     dashboard_ready = self.is_dashboard_ready()
>   File "/opt/app-root/lib64/python3.8/site-packages/codeflare_sdk/cluster/cluster.py", line 255, in is_dashboard_ready
>     response = requests.get(self.cluster_dashboard_uri(), timeout=5)
>   File "/opt/app-root/lib64/python3.8/site-packages/requests/api.py", line 73, in get
>     return request("get", url, params=params, **kwargs)
>   File "/opt/app-root/lib64/python3.8/site-packages/requests/api.py", line 59, in request
>     return session.request(method=method, url=url, **kwargs)
>   File "/opt/app-root/lib64/python3.8/site-packages/requests/sessions.py", line 573, in request
>     prep = self.prepare_request(req)
>   File "/opt/app-root/lib64/python3.8/site-packages/requests/sessions.py", line 484, in prepare_request
>     p.prepare(
>   File "/opt/app-root/lib64/python3.8/site-packages/requests/models.py", line 368, in prepare
>     self.prepare_url(url, params)
>   File "/opt/app-root/lib64/python3.8/site-packages/requests/models.py", line 439, in prepare_url
>     raise MissingSchema(
> requests.exceptions.MissingSchema: Invalid URL 'None': No scheme supplied. Perhaps you meant http://None?
> ```
> 
> Can you confirm whether SDK supports using Ingress? If so, can you check that Ingress is properly created when using KinD with Ingress installed in the test setup?

@sutaakar SDK not supporting Ingress yet.. Implementation is in progress for this https://github.com/project-codeflare/codeflare-sdk/pull/251

test/e2e/mnist_raycluster_sdk.py
```diff
@@ -18,11 +18,14 @@
     min_memory=0.5,
     max_memory=1,
     num_gpus=0,
+    image="quay.io/project-codeflare/ray:2.5.0-py38-cu116",
```
sutaakar, line 21: Why do you use custom image in here?

Srihari1192: I took same reference from the demo notebooks examples(https://github.com/project-codeflare/codeflare-sdk/blob/main/demo-notebooks/guided-demos/0_basic_ray.ipynb) assuming this is necessary  but looks this change is unnecessary the tests are pass now in local without this change,  earlier i am getting some error

test/e2e/olm_upgrade_test.go
```diff
@@ -0,0 +1,322 @@
...
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	. "github.com/project-codeflare/codeflare-operator/test/support"
+)
+
+var (
+	namespace          *corev1.Namespace
```
sutaakar, line 36: This is not needed.

test/e2e/olm_upgrade_test.go
```diff
@@ -0,0 +1,322 @@
...
+	namespaceName      string = "test-ns-olmupgrade"
+	serviceAccountName string = "sdk-user"
```
sutaakar, line 38:
```suggestion
	namespaceName       = "test-ns-olmupgrade"
	serviceAccountName = "sdk-user"
```

test/e2e/olm_upgrade_test.go
```diff
@@ -0,0 +1,322 @@
...
+)
+
+// Creates a Ray cluster
+func TestMNISTRayClusterUp(t *testing.T) {
+
+	test := With(t)
+	test.T().Parallel()
+	if os.Getenv("RUN_OLM_TESTS") != "true" {
```
sutaakar, line 46: It may be better to use build tags - https://stackoverflow.com/questions/54165975/go-test-only-run-tests-that-contain-a-build-tag
That way you can specify the OLMupgrade tests when invoking the tests, i.e. `go test -tags olm_upgrade_test ...`

test/e2e/olm_upgrade_test.go
```diff
@@ -0,0 +1,322 @@
...
+			APIVersion: corev1.SchemeGroupVersion.String(),
+			Kind:       "ConfigMap",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "mnist-raycluster-sdk",
+			Namespace: namespace.Name,
+		},
+		BinaryData: map[string][]byte{
+			// SDK script
+			"raycluster_sdk.py": ReadFile(test, "raycluster_sdk.py"),
```
sutaakar, line 72: The file name is not much descriptive.
Thinking about something like `start_ray_cluster.py`. What do you think?

test/e2e/olm_upgrade_test.go
```diff
@@ -0,0 +1,322 @@
...
+				},
+			},
+		},
+	}
+	job, err = test.Client().Core().BatchV1().Jobs(namespace.Name).Create(test.Ctx(), job, metav1.CreateOptions{})
+	test.Expect(err).NotTo(HaveOccurred())
+	test.T().Logf("Created Job %s/%s successfully", job.Namespace, job.Name)
+
+	// Retrieving the job logs once it has completed or timed out
+	defer WriteJobLogs(test, job.Namespace, job.Name)
```
sutaakar, line 202: This can be removed due to https://github.com/project-codeflare/codeflare-operator/pull/301

test/e2e/olm_upgrade_test.go
```diff
@@ -0,0 +1,322 @@
...
+		To(WithTransform(ConditionStatus(batchv1.JobComplete), Equal(corev1.ConditionTrue)))
+
+}
+
+// Submit a Job to the Ray cluster and trains the MNIST dataset using the CodeFlare SDK.
+func TestMnistJobSubmit(t *testing.T) {
+
+	test := With(t)
+	test.T().Parallel()
+	if os.Getenv("RUN_OLM_TESTS") != "true" {
```
sutaakar, line 222: Same here.

test/e2e/olm_upgrade_test.go
```diff
@@ -0,0 +1,322 @@
...
+				},
+			},
+		},
+	}
+	job, err = test.Client().Core().BatchV1().Jobs(namespaceName).Create(test.Ctx(), job, metav1.CreateOptions{})
+	test.Expect(err).NotTo(HaveOccurred())
+	test.T().Logf("Created Job %s/%s successfully", job.Namespace, job.Name)
+
+	// Retrieving the job logs once it has completed or timed out
+	defer WriteJobLogs(test, job.Namespace, job.Name)
```
sutaakar, line 309: Same here.

test/support/namespace.go
```diff
@@ -53,3 +53,34 @@ func deleteTestNamespace(t Test, namespace *corev1.Namespace) {
...
+
+	namespace, err := t.Client().Core().CoreV1().Namespaces().Create(t.Ctx(), namespace, metav1.CreateOptions{})
+	t.Expect(err).NotTo(gomega.HaveOccurred())
+
+	return namespace
+}
+
+func DeleteTestNamespace(t Test, namespaceName string) {
+	t.T().Helper()
+	propagationPolicy := metav1.DeletePropagationBackground
```
sutaakar, line 81: Can you add here:
```
	storeAllPodLogs(t, namespace)
	storeEvents(t, namespace)
```
after you rebase on latest main?

test/e2e/olm_upgrade_test.go
Srihari1192: yes sounds good

test/e2e/olm_upgrade_test.go
sutaakar: The disadvantage of this approach is that the file won't compile if the tag is not enabled.....
Thinking whether it may be better just to remove the condition and specify what tests to run in makefile - to have a separate command there to run upgrade tests.

Srihari1192: Yes also if we are adding tags, need to adjust to all the e2e tests with build tag to skip these tests running as part of e2e

Srihari1192: @sutaakar Probably we can use Test grouping for e2e test run like `go test -timeout 30m -v ./test/e2e -run "^TestMNIST.*$"` as all our e2e tests starts with TestMNIST and rename OLM upgrade test to **TestOLMUpgradeRayClusterUp** and **TestOLMUpgradeMnistJobSubmit** . So that we can call these tests specifically in our workflows by removing the condition

sutaakar: Or maybe this test can be moved to dedicated folder, i.e. `test/upgrade`.

Srihari1192: Sure will go with this approach

Srihari1192: Moved the tests to folder `test/upgrade`.. kept test dependent files in the `test/e2e` as [ReadFile](https://github.com/project-codeflare/codeflare-operator/blob/3bf1ba8cfa31872a28fac26c66914e1551fef94f/test/e2e/support.go#L30) method excepts files to be in the same package

test/e2e/olm_upgrade_test.go
Srihari1192: Done

test/support/namespace.go
Srihari1192: Added method `StoreNamespaceLogs`  to reuse in multiple places as we are not going to delete namespace when the test `TestMNISTRayClusterUp` Passed

sutaakar: a nitpick: what about using Namespace struct instead of namespaceName string? That way you don't need `GetNamespaceWithName` in `StoreNamespaceLogs`.

test/e2e/olm_upgrade_test.go
sutaakar: Thinking whether it would have sense to copy the method (doesn't have to be exported) to the `upgrade` package.

test/support/namespace.go
Srihari1192: Yes its a good idea, I will modify accordingly

test/e2e/olm_upgrade_test.go
Srihari1192: okay left as same as existing

test/support/namespace.go
Srihari1192: Added Namespace struct to initialise Namespace name and namespace instance

test/upgrade/olm_upgrade_test.go
```diff
@@ -0,0 +1,315 @@
...
+
+// Creates a Ray cluster
+func TestMNISTRayClusterUp(t *testing.T) {
+
+	test := With(t)
+	test.T().Parallel()
+
+	// Create a namespace
+	ns := NewNamespace(namespaceName)
+	namespace := ns.CreateTestNamespaceWithName(test, namespaceName)
```
sutaakar, line 48: Is there any reason for having to declare the `Namespace` struct?
Wouldn't it be enough to pass only namespace name, which is already declared in this file?
i.e.:
```suggestion
	namespace := CreateTestNamespaceWithName(test, namespaceName)
```

Srihari1192: Yes its not required here , i refactored to declare the Namespace struct for the test TestMnistJobSubmit and adjusted here aswell..

sutaakar: yes, but even in the `TestMnistJobSubmit` you can use just the namespace name declared as namespaceName variable in this file:
`namespace := GetNamespaceWithName(test, namespaceName)`
there is no need to create a dedicated struct for namespace as all the information you need you already have here

Srihari1192: Yeah right, will remove struct and update the same

sutaakar (commented):
/lgtm

test/upgrade/olm_upgrade_test.go
```diff
@@ -0,0 +1,206 @@
...
+	namespace := CreateTestNamespaceWithName(test, namespaceName)
+
+	// Delete namespace only if test failed
+	defer func() {
+		if t.Failed() {
+			DeleteTestNamespace(test, namespace)
+		} else {
+			StoreNamespaceLogs(test, namespace)
+		}
+	}()
```
astefanutti, line 54: @Srihari1192 @sutaakar out of curiosity, why not using the "standard" way, where test support does that automatically?

Srihari1192: @astefanutti In this Upgrade context, we are using the same namespace in after operator upgrade test.  As **NewTestNamespace** will delete the namespace by default after test complete , so we added this supported methods in [codeflare-common ](https://github.com/project-codeflare/codeflare-common/blob/988ba1da9069e1d8919da11761227daee6bce552/support/namespace.go#L57)

sutaakar: > It may be better to implement the upgrade as part of the Go test

This will couple the test with specific upgrade strategy (using OLM, overriding existing deployment with new oneliner, upgrade using ODH). Personally I would prefer keep test implementation aside from deployment/upgrade, to keep the test reusable for any strategy.

astefanutti: It makes sense. Sounds good 👍🏼.

PR #287, opened 2023-09-14
[262] Add links to other release notes

### LINKED ISSUE #262: Update Project CodeFlare Release Action to link to other release notes
Updating https://github.com/project-codeflare/codeflare-operator/blob/main/.github/workflows/project-codeflare-release.yml so that it includes link to release notes for all the sub-components would make the release notes much easier to parse for our users. Instead of having to dig through to find release notes, they can just click on a link

jiripetrlik:
# Issue link
Closes #262 

Tested in https://github.com/jiripetrlik/codeflare-operator/actions/runs/6184331678/job/16787779852 . However, I removed all not necessary steps.

### DIFF
--- a/.github/workflows/tag-and-build.yml
+++ b/.github/workflows/tag-and-build.yml
@@ -85,10 +85,10 @@ jobs:
 
     - name: Adjust Compatibility Matrix in readme
       run: |
-        sed -i -E "s/(.*CodeFlare Operator.*)v[0-9]+\.[0-9]+\.[0-9]+(.*)/\1${{ github.event.inputs.version }}\2/" README.md
-        sed -i -E "s/(.*Multi-Cluster App Dispatcher.*)v[0-9]+\.[0-9]+\.[0-9]+(.*)/\1${{ github.event.inputs.mcad-version }}\2/" README.md
-        sed -i -E "s/(.*CodeFlare-SDK.*)v[0-9]+\.[0-9]+\.[0-9]+(.*)/\1${{ github.event.inputs.codeflare-sdk-version }}\2/" README.md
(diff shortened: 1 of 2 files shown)


.github/workflows/tag-and-build.yml
```diff
@@ -84,6 +84,13 @@ jobs:
       run: make install-operator-sdk
 
     - name: Adjust Compatibility Matrix in readme
+      run: |
+        sed -i -E "s/(.*CodeFlare Operator.*)v[0-9]+\.[0-9]+\.[0-9]+(.*\|.*\|)/\1${{ github.event.inputs.version }}\2/" README.md
```
sutaakar, line 88: There is no need for the regex alteration.
If you call the regex below twice it will replace both values.
Not sure if there is a possibility to do both replacements in one command.

sutaakar:
Thinking whether it would have sense to put the link directly into the version value in readme instead of extra column, though I don't have a strong opinion on that.

.github/workflows/tag-and-build.yml
KPostOffice: ```sed -i -E "s/(.*CodeFlare Operator.*)v[0-9]+\.[0-9]+\.[0-9]+(.*)v[0-9]+\.[0-9]+\.[0-9]+(.*)/\1${{ github.event.inputs.version }}\2${{ github.event.inputs.version }}\3/" README.md```

Single line sed, I tested this in my local repo and it seems to work well.

jiripetrlik:
@sutaakar @KPostOffice 
Thank you for valuable comments. It should be changed. Test run can be found here: https://github.com/jiripetrlik/codeflare-operator/actions/runs/6223842551/job/16890708698

PR #304, opened 2023-09-22
refactor: addition of rbac needed for instascale controller

### LINKED ISSUE #278: Add RBAC required by the InstaScale controller
## WHY

Part of the CodeFlare operator re-design, as described in project-codeflare/adr#9. 

## WHAT

Include the RBAC manifests required by the InstaScale controller, e.g., those required to consume the MachineSet and ClusterVersion APIs. 

## HOW

Copy the InstaScale RBAC into separate manifests.

To be done on top of #216.

## TESTS

Make sure the embedded InstaScale controller works as expected.

dimakis:
# Issue link

closes #278 

# What changes have been made

Addition of a ClusterRole and a ClusterRoleBinding which binds the required permissions to the CFO serviceaccount

# Verification steps

Run through demo instascale NBs with clusters leveraging machinepools (OSD) and machinesets (instructions for sample can be found [here](https://github.com/Maxusmusti/dev-clusters))

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/config/rbac/instascale_role.yaml
+++ b/config/rbac/instascale_role.yaml
@@ -8,12 +8,6 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - configmaps
-  verbs:
-  - get
-- apiGroups:
-  - ""
-  resources:
   - nodes
   verbs:
   - get
@@ -22,20 +16,10 @@ rules:
   - update
 - apiGroups:
   - ""
-  resourceNames:
-  - instascale-ocm-secret
   resources:
   - secrets
   verbs:
   - get
-- apiGroups:
-  - apps
-  resources:
-  - deployments
-  verbs:
-  - get
-  - list
-  - watch
 - apiGroups:
   - apps
   resources:
@@ -65,7 +49,10 @@ rules:
 - apiGroups:
   - machine.openshift.io
   resources:
-  - '*'
+  - controlplanemachinesets
+  - machinehealthchecks
+  - machines
+  - machinesets
   verbs:
   - create
   - delete
@@ -74,29 +61,3 @@ rules:
   - patch
   - update
   - watch
-- apiGroups:
-  - workload.codeflare.dev
-  resources:
-  - appwrappers
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
-- apiGroups:
-  - workload.codeflare.dev
-  resources:
-  - appwrappers/finalizers
-  verbs:
-  - update
-- apiGroups:
-  - workload.codeflare.dev
-  resources:
-  - appwrappers/status
-  verbs:
-  - get
-  - patch
-  - update
--- a/config/rbac/instascale_role.yaml
+++ b/config/rbac/instascale_role.yaml
@@ -21,24 +21,6 @@ rules:
   verbs:
   - get
 - apiGroups:
-  - apps
-  resources:
-  - machineset
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
-- apiGroups:
-  - apps
-  resources:
-  - machineset/status
-  verbs:
-  - get
-- apiGroups:
   - config.openshift.io
   resources:
   - clusterversions
--- a/config/rbac/instascale_role.yaml
+++ b/config/rbac/instascale_role.yaml
@@ -49,8 +49,6 @@ rules:
 - apiGroups:
   - machine.openshift.io
   resources:
-  - controlplanemachinesets
-  - machinehealthchecks
   - machines
   - machinesets
   verbs:
--- a/config/rbac/instascale_role.yaml
+++ b/config/rbac/instascale_role.yaml
@@ -1,0 +1,102 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  name: instascale-role
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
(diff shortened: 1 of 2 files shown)


config/rbac/instascale_role.yaml
```diff
@@ -0,0 +1,102 @@
...
+  name: instascale-role
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+- apiGroups:
+  - ""
```
astefanutti, line 15: It's already part of the "core" CusterRole, so it's preferable not duplicating it.

config/rbac/instascale_role.yaml
```diff
@@ -0,0 +1,102 @@
...
+  verbs:
+  - get
+- apiGroups:
+  - apps
+  resources:
+  - deployments
+  verbs:
+  - get
+  - list
+  - watch
```
astefanutti, line 38: It's already part of the main/MCAD CusterRole, so it's preferable not duplicating it.

config/rbac/instascale_role.yaml
```diff
@@ -0,0 +1,102 @@
...
+  resources:
+  - clusterversions
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - machine.openshift.io
+  resources:
+  - '*'
```
astefanutti, line 68: Could wildcard be avoided? It's generally a best practice recommended by SRE.

config/rbac/instascale_role.yaml
```diff
@@ -0,0 +1,102 @@
...
+  - nodes
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+- apiGroups:
+  - ""
+  resourceNames:
+  - instascale-ocm-secret
```
astefanutti, line 26: The OCM Secret can now be referenced in the operator configuration, so it's not hard-coded anymore.

config/rbac/instascale_role.yaml
```diff
@@ -0,0 +1,102 @@
...
+  verbs:
+  - update
+- apiGroups:
+  - workload.codeflare.dev
+  resources:
+  - appwrappers/status
+  verbs:
+  - get
+  - patch
+  - update
```
astefanutti, line 102: It's already part of the main/MCAD CusterRole, so it's preferable not duplicating these.

astefanutti (changes requested):

config/rbac/instascale_role.yaml
```diff
@@ -65,19 +49,10 @@ rules:
...
-  - list
-  - patch
-  - update
-  - watch
-- apiGroups:
-  - workload.codeflare.dev
-  resources:
-  - appwrappers
+  - controlplanemachinesets
+  - machinehealthchecks
```
astefanutti, line 53: Are these two resources really used by InstaScale?

dimakis: I should have specified I've pushed this as myself and mark are testing which are actually used and what we can cut away.

astefanutti: My understanding is that InstaScale only uses the Machine and MachineSet APIs but better double checking.

Bobbins228 (changes requested):
Tested this on an OCP cluster with Antonin's requested changes and got expected InstaScale behaviour.

When testing with OSD I received this error on scale up:
```
E0925 15:58:54.827990 1 machinepools.go:58] Error creating MachinePool: expected response content type 'application/json' but received '' and content ''
I0925 15:58:54.828009 1 machinepools.go:60] Created MachinePool: <nil>
E0925 15:58:55.263876 1 machinepools.go:58] Error creating MachinePool: expected response content type 'application/json' but received '' and content ''
I0925 15:58:55.263893 1 machinepools.go:60] Created MachinePool: <nil>
```

astefanutti:
> When testing with OSD I received this error on scale up:
> 
> ```
> E0925 15:58:54.827990 1 machinepools.go:58] Error creating MachinePool: expected response content type 'application/json' but received '' and content ''
> I0925 15:58:54.828009 1 machinepools.go:60] Created MachinePool: <nil>
> E0925 15:58:55.263876 1 machinepools.go:58] Error creating MachinePool: expected response content type 'application/json' but received '' and content ''
> I0925 15:58:55.263893 1 machinepools.go:60] Created MachinePool: <nil>
> ```

Could it be that we are missing permissions on `machinepools` or other related resources?

config/rbac/instascale_role.yaml
```diff
@@ -0,0 +1,61 @@
...
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - apps
+  resources:
+  - machineset/status
+  verbs:
+  - get
```
astefanutti, line 40: Are MachineSet APIs really in the `apps` group? It seems it's a wrong duplicate of the `machine.openshift.io` one.

dimakis: nope, I didn't think they were in the `apps` group, I thought first class objects like replicasets, deployments etc were.

but I was just copying what was in the [original role](https://github.com/project-codeflare/instascale/blob/2c7538eb8f93b77a768bd774bf963f37a0c41b23/config/rbac/role.yaml#L39-L56) to try get it working.

I can't really figure out why this is the case, even getting rid of those and allowing the wildcard on teh machine.openshift.io group doesn't returns the same error as ye got. 
I also can't seem to find a machinepool resource for any api group in any of the docs. 
Any ideas here?

astefanutti: Actually, I don't think the error is related to the RBAC changes. I suspect there is an issue with the OCM client configuration.

PR #306, opened 2023-09-25
[263] Release Action can skip intermediate steps

### LINKED ISSUE #263: Update Release Action to support skipping of intermediate steps
Updating https://github.com/project-codeflare/codeflare-operator/blob/main/.github/workflows/project-codeflare-release.yml to allow us to skip components which don't require new releases would help simplify the release process. Ideally, the release action automatically skips a component if a release with the specified version already exists.

jiripetrlik:
Closes #263 
It was necessary to use only one additional parameter for the list of released components because Github allows only 10 workflow parameters. I was also forced to use "if" for steps instead of jobs. It is because if I skip the job then all dependent jobs are also skipped which is not desirable.

Example runs (all bash code was commented out to not perform real release):
- https://github.com/jiripetrlik/codeflare-operator/actions/runs/6300310513
- https://github.com/jiripetrlik/codeflare-operator/actions/runs/6300326727

### DIFF
--- a/.github/workflows/project-codeflare-release.yml
+++ b/.github/workflows/project-codeflare-release.yml
@@ -41,12 +41,21 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - name: Check if MCAD release does exist
+      run: |
+        status_code=$(curl -s -o /dev/null -w "%{http_code}" https://github.com/project-codeflare/multi-cluster-app-dispatcher/releases/tag/${{ github.event.inputs.mcad-version }})
+        if [[ "$status_code" == "200" ]]; then
+          echo "MCAD release with version ${{ github.event.inputs.mcad-version }} already exist. Will not create MCAD release."
+        fi
+        echo "MCAD_RELEASE_STATUS_CODE=$status_code" >> $GITHUB_ENV
+
     - name: Release MCAD
       run: |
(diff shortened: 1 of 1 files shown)


sutaakar:
@jiripetrlik What about using a different logic here - if component version is not defined then we can consider component build to be skipped?
That would also require to adjust the release workflow in components to make their dependencies optional too.

What do you think about it?

jiripetrlik:
@sutaakar
I've changed it in the way we discussed yesterday. Release of component will be triggered only if it doesn't exist yet.

Test runs:
- https://github.com/jiripetrlik/codeflare-operator/actions/runs/6318726948
- https://github.com/jiripetrlik/codeflare-operator/actions/runs/6318818278
- https://github.com/jiripetrlik/codeflare-operator/actions/runs/6318841233

.github/workflows/project-codeflare-release.yml
```diff
@@ -41,12 +41,18 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - name: Check if MCAD release does exist
+      run: |
+        status_code=$(curl -s -o /dev/null -w "%{http_code}" https://github.com/project-codeflare/multi-cluster-app-dispatcher/releases/tag/${{ github.event.inputs.mcad-version }})
```
sutaakar, line 46: Have you considered using GH CLI?
https://github.com/project-codeflare/codeflare-operator/blob/main/.github/workflows/tag-and-build.yml#L65-L75

It is longer but may be more descriptive.
Also it would be good to notify users that the release already exists.

anishasthana (approved):
/lgtm

PR #312, opened 2023-09-29
fix: codeflare-operator e2e tests not passing

VanillaSpoon:
# Issue link
#313

# What changes have been made
Using the 'PYTHONUSERBASE' environment variable, to point to the emptyDir volume mount for mnist_pytorch_mcad_job_test and mnist_raycluster_sdk_test

Updated the readme to provide instructions to setting the CODEFLARE_TEST_OUTPUT_DIR when running e2e tests locally.

# Verification steps
Follow the end to end test instructions, and they should now pass, and complete the runthrough :)

## Checks
- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/test/e2e/mnist_pytorch_mcad_job_test.go
+++ b/test/e2e/mnist_pytorch_mcad_job_test.go
@@ -77,15 +77,23 @@ func TestMNISTPyTorchMCAD(t *testing.T) {
 				Spec: corev1.PodSpec{
 					Containers: []corev1.Container{
 						{
-							Name:    "job",
-							Image:   GetPyTorchImage(),
+							Name:  "job",
+							Image: GetPyTorchImage(),
+							Env: []corev1.EnvVar{
+								corev1.EnvVar{Name: "PYTHONUSERBASE", Value: "/workdir"},
+							},
 							Command: []string{"/bin/sh", "-c", "pip install -r /test/requirements.txt && torchrun /test/mnist.py"},
 							VolumeMounts: []corev1.VolumeMount{
 								{
 									Name:      "test",
 									MountPath: "/test",
 								},
+								{
+									Name:      "workdir",
+									MountPath: "/workdir",
+								},
 							},
+							WorkingDir: "/workdir",
 						},
 					},
 					Volumes: []corev1.Volume{
@@ -97,6 +105,12 @@ func TestMNISTPyTorchMCAD(t *testing.T) {
 										Name: config.Name,
 									},
 								},
+							},
+						},
+						{
+							Name: "workdir",
+							VolumeSource: corev1.VolumeSource{
+								EmptyDir: &corev1.EmptyDirVolumeSource{},
 							},
 						},
 					},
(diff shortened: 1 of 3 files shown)


test/support/utils.go
```diff
@@ -36,6 +36,24 @@ const (
 
 func WriteToOutputDir(t Test, fileName string, fileType OutputType, data []byte) {
 	t.T().Helper()
-	t.Expect(os.WriteFile(path.Join(t.OutputDir(), fileName+"."+string(fileType)), data, fs.ModePerm)).
-		To(gomega.Succeed())
+
+	// Get output directory and create the directory if it doesn't exist
+	outputDir := t.OutputDir()
+	err := os.MkdirAll(outputDir, os.ModePerm)
```
astefanutti, line 42: If we _really_ want to have the parent directory created, it'd be better to encapsulate the logic in the `OutputDir()` method. Still the existing assumption is that the parent directory exist, and the output directory is created there.

test/e2e/mnist_pytorch_mcad_job_test.go
```diff
@@ -77,15 +77,23 @@ func TestMNISTPyTorchMCAD(t *testing.T) {
...
 								{
 									Name:      "test",
 									MountPath: "/test",
 								},
+								{
+									Name:      "test2",
+									MountPath: "/test2",
+								},
 							},
+							WorkingDir: "/test2",
```
astefanutti, line 96: Can we find a better name for the volume and the directory, maybe `workdir`?

test/e2e/mnist_raycluster_sdk_test.go
```diff
@@ -160,14 +160,22 @@ func TestMNISTRayClusterSDK(t *testing.T) {
...
 								{
 									Name:      "test",
 									MountPath: "/test",
 								},
+								{
+									Name:      "test2",
+									MountPath: "/test2",
+								},
 							},
+							WorkingDir: "/test2",
```
astefanutti, line 178: Can we find a better name for the volume and the directory, maybe `workdir`?

test/support/utils.go
VanillaSpoon: Thanks @astefanutti ,
For some reason the /tmp/TestMNISTRayJobMCADRayCluster directory isn't being created before that, so the logs couldn't be saved.

test/e2e/mnist_pytorch_mcad_job_test.go
VanillaSpoon: Thanks @astefanutti , I'll update this :)

test/support/utils.go
astefanutti: In what scenario did it fail? Is it when you run the test locally?

VanillaSpoon: Yea, I've been running the codeflare-operator locally via openshift

astefanutti: OK, so when run locally, the `CODEFLARE_TEST_OUTPUT_DIR` environment variable is supposed to be used. It's somehow designed so as to avoid having the test execution polluting the disk by mistake.

VanillaSpoon: Very Nice :D Thanks @astefanutti  these changes have been pushed.

astefanutti:
Thanks!

/lgtm

PR #325, opened 2023-10-09
Fixing type in ocm secret name

Fiona-Waters:
Fixing a type following on from [Instascale e2e PR](https://github.com/project-codeflare/codeflare-operator/pull/271).


sutaakar (approved):
/lgtm

Issue #366, opened 2023-10-23
Fix GH action to trigger sync job in ODH


dimakis:
This [GH action ](https://github.com/project-codeflare/codeflare-operator/actions/workflows/odh-fork-sync.yml) failed to run during the last release of the CF stack. The action had to be run manually. 

This needs to be fixed

dimakis:
cc @Bobbins228

PR #373, opened 2023-10-24
Updated CFO sync and tag-build workflows

Bobbins228:
# Issue link

Closes: #366 
# What changes have been made

During the release on Friday the `odh-fork-sync` workflow did not trigger. I suspect the issue is similar to this one [here](https://github.com/orgs/community/discussions/26875#discussioncomment-3253761) . 
I updated the `tag-and-build` workflow to use the `CODEFLARE_MACHINE_ACCOUNT_TOKEN` instead of the `github.TOKEN` which could solve our issue.

I also made changes to how `odh-sync-fork` is set up.
# Verification steps

I tested this out using example workflows and I got a workflow to trigger on a published release using the PAT token.
## Checks
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/.github/workflows/odh-fork-sync.yml
+++ b/.github/workflows/odh-fork-sync.yml
@@ -11,7 +11,7 @@ jobs:
     steps:
       - name: Sync-fork
         run: |
-          gh workflow run sync-fork.yaml -R github.com/opendatahub-io/codeflare-operator -r main
+          gh workflow run sync-fork.yaml --repo github.com/opendatahub-io/codeflare-operator --ref main
         env:
           GITHUB_TOKEN: ${{ secrets.CODEFLARE_MACHINE_ACCOUNT_TOKEN }}
         shell:
--- a/.github/workflows/odh-fork-sync.yml
+++ b/.github/workflows/odh-fork-sync.yml
@@ -1,7 +1,7 @@
 name: Call sync on OpenDataHub CFO fork sync
 on:
   release:
-    types: 
+    types:
       [published]
   workflow_dispatch:
 
--- a/.github/workflows/odh-fork-sync.yml
+++ b/.github/workflows/odh-fork-sync.yml
@@ -1,7 +1,8 @@
 name: Call sync on OpenDataHub CFO fork sync
 on:
   release:
-    types: [published]
+    types: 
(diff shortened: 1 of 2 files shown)


.github/workflows/odh-fork-sync.yml
```diff
@@ -1,15 +1,18 @@
 name: Call sync on OpenDataHub CFO fork sync
 on:
   release:
-    types: [published]
+    types:
+      [published]
```
astefanutti, line 5: Shouldn't that be:
```
types:
  - published
```

Bobbins228: I am sure it could be but I just followed the [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release)

astefanutti: I thought it's either:
```
types: [published]
```
Or:
```
types:
  - published
```
But maybe:
```
types:
  [published]
```
Also works. Just that it's a bit uncommon :).

Bobbins228: All three of those work but to keep things simple Ill set to 
```
types:
  - published
```
As we are only working with 1 type not multiple so theres no need for a list :+1:

astefanutti:
@sutaakar could you have a quick look?

sutaakar:
looking....

.github/workflows/odh-fork-sync.yml
```diff
@@ -1,15 +1,18 @@
...
 
 jobs:
   sync-fork:
     runs-on: ubuntu-latest
     steps:
       - name: Sync-fork
-        env:
-          GITHUB_TOKEN: ${{ secrets.CODEFLARE_MACHINE_ACCOUNT_TOKEN }}
         run: |
           gh workflow run sync-fork.yaml -R github.com/opendatahub-io/codeflare-operator -r main
```
sutaakar, line 14: Just optional suggestion:
```suggestion
          gh workflow run sync-fork.yaml --repo github.com/opendatahub-io/codeflare-operator --ref main
```
From my perspective that improves readability of the command.

PR #374, opened 2023-10-24
Use kuberay/ray-operator@vX.Y.Z Go module

### LINKED ISSUE #354: Use kuberay/ray-operator@vX.Y.Z Go module
Until now, KubeRay has not been tagging Go modules in such a way a specific version can be consumed.

Once https://github.com/ray-project/kuberay/pull/1546 is applied, we should move to using the tagged version.

It may have to be done in codeflare-common once tests are moved into this repository after #336 is merged.

astefanutti:
Leverages https://github.com/ray-project/kuberay/pull/1546.

Closes #354.


sutaakar (commented):
/lgtm

sutaakar (approved):
/lgtm

PR #376, opened 2023-10-24
Added launch.json for local debugging with VSCode

Bobbins228:
# Issue link

# What changes have been made

Added launch.json file for local debugging of the CodeFlare Operator
# Verification steps

Fill in the launch.json file with your kube config file path and namespace the CFO should be acting in.
Scale down the CodeFlare Operator deployment if it exists in your cluster.
The operator should be running locally with VSCode.
## Checks
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,5 +30,5 @@ bundle.Dockerfile
 
 .DS_STORE
 
-#VSCode
+# VSCode
 .vscode
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@ bundle.Dockerfile
 *~
 
 .DS_STORE
+
+#VSCode
+.vscode
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,17 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+      {
+        "name": "Debug CodeFlare Operator",
+        "type": "go",
+        "request": "launch",
+        "mode": "auto",
+        "program": "${workspaceFolder}/main.go",
+        "env": {
+          "KUBECONFIG": "</path/to/your/kube/config>",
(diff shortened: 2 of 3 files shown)


tedhtchang (commented):
@Bobbins228 Could you add an example on how you run local debugger with the launch.json in the https://github.com/project-codeflare/codeflare-operator/blob/main/CONTRIBUTING.md

Fiona-Waters (commented):
/lgtm

KPostOffice (approved):
I was able to get a debugging session running with the given instructions. Thanks @Bobbins228

ronensc:
@Bobbins228 Thanks!
After I set `.vscode/launch.json` with my custom values, every time I run `git status` it appears as a modified file.
I was able to make git ignore it by:
```
git update-index --skip-worktree .vscode/launch.json
```
The reverse command is:
```
git update-index --no-skip-worktree .vscode/launch.json
```
ref: https://stackoverflow.com/a/50222853

Did you find a better solution to make git ignore this custom changes?

Bobbins228:
@ronensc That's odd I see what you mean. I included the `.vscode` folder in the `.gitignore` file so by all means it should be ignored when changes are added to `launch.json`

ronensc:
AFAIK, once git tracks a file, you can't ignore it with `.gitignore`.
An alternative approach would be to save the template file as `launch.json.example` and track it with git. Then, each developer can make a copy of it locally (`launch.json`) that will be added to `.gitignore`. Although, I'm not sure this approach is better.

Bobbins228:
I see I'll look into a way of handling this and if there isn't any other solution we'll go with an example file. Thanks for the feedback!

PR #390, opened 2023-11-01
Fix broken catalog-build-from-index command

sutaakar:
# Issue link

# What changes have been made

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -331,7 +331,7 @@ catalog-build-from-index: opm ## Build a catalog image.
 	mkdir catalog
 	$(OPM) render $(CATALOG_BASE_IMG) -o yaml > catalog/bundles.yaml
 	$(OPM) render $(BUNDLE_IMG) $(OPM_BUNDLE_OPT) > catalog/codeflare-operator-bundle.yaml
-	$(SED) -i -E "s/(.*)(- name: codeflare-operator.$(PREVIOUS_VERSION).*)/\1- name: codeflare-operator.$(VERSION)\n  replaces: codeflare-operator.$(PREVIOUS_VERSION)\n\2/" catalog/bundles.yaml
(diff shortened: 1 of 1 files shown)


Bobbins228 (approved):
/lgtm

Makefile
```diff
@@ -331,7 +331,7 @@ catalog-build-from-index: opm ## Build a catalog image.
 	mkdir catalog
 	$(OPM) render $(CATALOG_BASE_IMG) -o yaml > catalog/bundles.yaml
 	$(OPM) render $(BUNDLE_IMG) $(OPM_BUNDLE_OPT) > catalog/codeflare-operator-bundle.yaml
-	$(SED) -i -E "s/(.*)(- name: codeflare-operator.$(PREVIOUS_VERSION).*)/\1- name: codeflare-operator.$(VERSION)\n  replaces: codeflare-operator.$(PREVIOUS_VERSION)\n\2/" catalog/bundles.yaml
+	$(SED) -i -E "s/(.*)(- name: codeflare-operator.v0.2.0)/\1- name: codeflare-operator.$(VERSION)\n  replaces: codeflare-operator.$(PREVIOUS_VERSION)\n\2/" catalog/bundles.yaml
```
dimakis, line 334: excuse my lack of sed knowledge here but are we now hardcoding a version? is this PR just for this fix?

dimakis (commented):
just the one comment

Makefile
sutaakar: The purpose of this line is to inject replacing entry into the bundle section, where replacing entries are kept.
Originally this new replacing entry is placed right before last CodeFlare operator entry, however due to the regex issues with v1.0.0 the injection was duplicated. By fixing the searched text I made sure that only one injection is done.

Anyway this code is used just for tests, for real OLM release this bundle index is generated in OLM repo, we just provide manifests.

Issue #380, opened 2023-10-31
Update/ Add the action which creates the CF stack compatibility matrix


dimakis:
The KubeRay line in the compatibility matrix doesn't update. This needs to be fixed.

dimakis:
@ChughShilpa the readme compatibility matrix doesn't get updated when we create a new tag (here)[https://github.com/opendatahub-io/kuberay/tags] this should be updated via an action preferably

sutaakar:
@ChughShilpa can this issue be clodes?

astefanutti:
it seems the latest run did not update the Kuberay version correctly: https://github.com/project-codeflare/codeflare-operator/pull/407/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R15.

sutaakar:
Good catch, the parameter is missing in https://github.com/project-codeflare/codeflare-operator/blob/main/.github/workflows/project-codeflare-release.yml

sutaakar:
Fix ready ^

PR #397, opened 2023-11-07
Update workflow to add kuberay in compatibility matrix

ChughShilpa:
# Issue link

Closes #<[380](https://github.com/project-codeflare/codeflare-operator/issues/380)>

# What changes have been made
Updating `Tag and build`GH action to add KubeRay line in the compatibilty matrix

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/.github/workflows/tag-and-build.yml
+++ b/.github/workflows/tag-and-build.yml
@@ -25,7 +25,7 @@ on:
         required: true
         default: 'v0.0.0-dev'
       kuberay-version:
-        description: 'Published version of KubeRay'
+        description: 'Tested version of KubeRay'
         required: true
         default: 'v0.0.0-dev'
       is-stable:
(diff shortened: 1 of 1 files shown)


.github/workflows/tag-and-build.yml
```diff
@@ -24,6 +24,10 @@ on:
         description: 'Published version of InstaScale'
         required: true
         default: 'v0.0.0-dev'
+      kuberay-version:
+        description: 'Published version of KubeRay'
```
astefanutti, line 28: We don't publish a KubeRay version strictly speaking. So maybe we can describe it as "Tested version of KubeRay"?

ChughShilpa: Updated the description

sutaakar (approved):
/lgtm

PR #398, opened 2023-11-08
Added Workflow job to update CFO image

### LINKED ISSUE #392: Update automation to update the image version in `params.env`
[This value](https://github.com/opendatahub-io/codeflare-operator/blob/10db24421408a0f8f64c99690b2d106acd703015/config/manifests/bases/params.env#L2) needs to be updated with the correct version in the release action

[This commit](https://github.com/opendatahub-io/codeflare-operator/commit/97facffce2adb8fc7c3680ef669d35c161de0e05) may need to be replicated in the PCF-CFO as it does not currently exist in the upstream

cc @anishasthana

Bobbins228:
# Issue link

Closes #392 
# What changes have been made

Added a job to the release workflow that will update the version of the CFO image to the latest version.
Commit job has been updated to commit `*.env` files
Updated `config/manifests/kustomization.yaml` and added image stream files
# Verification steps

Test the perl command for updating the image locally
Run `make deploy -e IMG=<YOUR_DEV_IMAGE> -e ENV="e2e"`  for Kubernetes
Run `make deploy -e IMG=<YOUR_DEV_IMAGE>` for OpenShift
## Checks
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests


config/manifests/bases/params.yaml
```diff
@@ -0,0 +1,5 @@
+varReference:
+  - path: subjects[]/namespace
+    kind: ClusterRoleBinding
```
VanillaSpoon, line 3: I believe this is no longer necessary

```diff
-  - path: subjects[]/namespace
-    kind: ClusterRoleBinding
```

config/manifests/bases/params.env
```diff
@@ -0,0 +1,2 @@
+namespace=opendatahub
+odh-codeflare-operator-controller-image=quay.io/project-codeflare/codeflare-operator:v1.0.0
```
VanillaSpoon, line 2: Namespace can also be removed as it is not longer necessary for the CRB.

config/manifests/kustomization.yaml
```diff
@@ -1,8 +1,28 @@
...
+    apiVersion: v1
+  fieldref:
+    fieldpath: data.namespace
+- name: odh_codeflare_operator_controller_image
+  objref:
+    kind: ConfigMap
+    name: codeflare-stack-config
+    apiVersion: v1
+  fieldref:
+    fieldpath: data.odh-codeflare-operator-controller-image
```
VanillaSpoon, line 25: The `odh-codeflare-operator-controller-image`  is being updated correctly for the version, and added to the configmap correctly, but it is not being set in the deployment anywhere, so we will need to adjust [here](https://github.com/project-codeflare/codeflare-operator/blob/main/config/manager/manager.yaml#L37) to use the image: 

```diff
-        image: controller:latest
+        image: $(odh_codeflare_operator_controller_image)

```


Namespace can also be removed from here

```diff
- - name: namespace
-  objref:
-    kind: ConfigMap
-    name: codeflare-stack-config
-    apiVersion: v1
-  fieldref:
-    fieldpath: data.namespace
```

VanillaSpoon:
Other than those small adjustments, lgtm. Real nifty change to have it automatically update :)

astefanutti (commented):
We may have to adapt the way the image is overwritten in `make deploy` so it works and the e2e tests pass.

VanillaSpoon:
Hey @Bobbins228 

The tests are failing due to kustomize running from default, see  [ENV](https://github.com/project-codeflare/codeflare-operator/blob/main/Makefile#L86C8-L86C8) variable in the makefile. 

When running make deploy, this ensures the [build](https://github.com/project-codeflare/codeflare-operator/blob/main/Makefile#L183C1-L183C1) from kustomize is from within the default directory. But with the updates in the pr, you're changes are within /config/manifests, which means for those to update the deployment the ENV would need to be set to manifests.

To see this you can run `kustomize build` from each of those directories, and see the image updated from building within config/manifests.

Makefile
```diff
@@ -83,7 +83,7 @@ ENVTEST_K8S_VERSION = 1.24.2
 
 # The target deployment environment, that corresponds to the Kustomize directory
 # used to build the manifests.
-ENV ?= default
+ENV ?= manifests
```
astefanutti, line 86: Could you please make sure `make deploy` takes the image into account correctly with `cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}`. It may be the tests pass because the default image is used, while it should be using the locally built one.

Bobbins228: It seems there's a conflict.
While we have the image set to `$(odh-codeflare-operator-controller-image)`  in `manager.yaml` Kustomize cannot overwrite it with the passed image because you can't use the variable in the command e.g. 
`$(KUSTOMIZE) edit set image $(odh-codeflare-operator-controller-image)=${IMG}`.

We could revert that variable back to `controller:latest` and alter the make file so that if there was no image passed it would default to the latest image in params.env. wdyt?

astefanutti: Would replacing `$(KUSTOMIZE) edit set image $(odh-codeflare-operator-controller-image)=${IMG}` with something that changes the value in `params.env` with the value of `IMG` be option?

Bobbins228: Absolutely we can use a similar perl command from the workflow portion. I'll get on that. Thanks!

astefanutti:
@sutaakar you OK with using `perl`?

sutaakar:
It is ok for me.
Just thinking about the Notebook ImageStream, whether it may be better to keep it just in ODH/CFO.

astefanutti:
> It is ok for me. Just thinking about the Notebook ImageStream, whether it may be better to keep it just in ODH/CFO.

Right, I was actually surprise the deployment in the e2e tests pass, which made me realise it uses the `e2e` overlay, that depends on the `default` one, which by-passes that new layer. We should make sure that new one is exercise by the tests. I'll also be used by the one-liner installation.

astefanutti:
@Bobbins228 @sutaakar Or we accept that `manifests` overlay is only for ODH.

sutaakar:
I would prefer having manifest in `ODH/CFO` though no strong opinion about that.

config/manifests/bases/params.env
```diff
@@ -0,0 +1 @@
+odh-codeflare-operator-controller-image=quay.io/opendatahub/codeflare-operator:v1.0.0
```
sutaakar, line 1: @astefanutti Do you know who/when/how builds `quay.io/opendatahub/codeflare-operator`? Should we build it?

astefanutti: @sutaakar ideally it should be build from the ODH/CFO fork, by the same same workflow with have here in PCF/CFO, just changing the image org?

Bobbins228: We have a workflow for building this image in odh/cfo [here](https://github.com/opendatahub-io/codeflare-operator/blob/main/.github/workflows/build-and-push.yaml)

sutaakar: Right, I just overlook it. Thanks.

sutaakar (commented):
/lgtm

config/manifests/bases/params.env
sutaakar: In general the flow is not ideal as CFO image here is referenced sooner than it is actually built.

config/manifests/kustomization.yaml
```diff
@@ -6,3 +6,19 @@ resources:
 - ../scorecard
 apiVersion: kustomize.config.k8s.io/v1beta1
```
astefanutti, line 7: I think we should remove the CSV file and its reference there? @anishasthana @sutaakar can you confirm?

sutaakar: I think so

Bobbins228: The CSV file?

astefanutti: The `bases/codeflare-operator.clusterserviceversion.yaml` file.

config/manifests/kustomization.yaml
```diff
@@ -1,8 +1,23 @@
...
-- bases/codeflare-operator.clusterserviceversion.yaml
 - ../default
 - ../scorecard
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
+
+configMapGenerator:
+- name: codeflare-stack-config
+  envs:
+    - bases/params.env
```
astefanutti, line 12: I'd suggest we take that opportunity to move these two files in the `manifests` directory and remove the `bases` one. It's a leftover from the operator SDK for how it handles the CSV file, and it's not relevant anymore.

config/manifests/kustomization.yaml
astefanutti: Ah so I forgot we still run the OLM upgrade test. @sutaakar what's your opinion on this? Do we want to keep them around, in which case we need to keep the CSV?

Bobbins228:
@sutaakar @astefanutti It seems after removing the `bases/codeflare-operator.clusterserviceversion.yaml` a new one is created through this Make [command](https://github.com/project-codeflare/codeflare-operator/pull/398/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L286). 

It has a few mandatory fields that must be passed which is causing the OLM install to fail. Do we need to generate this anymore since we chose to remove `bases/codeflare-operator.clusterserviceversion.yaml`?

config/manifests/kustomization.yaml
```diff
@@ -3,6 +3,7 @@
 resources:
 - ../default
 - ../scorecard
+- codeflare-operator.clusterserviceversion.yaml
```
astefanutti, line 6: We have to find a way to keep the file, but not have it in the manifests overlay, otherwise it'll be taken into account by the ODH operator.

Bobbins228: I don't think there is a way to run the `make bundle` command without the CSV in the kustomization file which is what causes the spec.Icon error in the previous commit OLM workflow.

astefanutti: Right, now that the `manifests` directory is used by the ODH operator, we need to move the CSV file into a separate directory / overlay. WDYT?

Bobbins228: I can test this out :+1:

Bobbins228: It seems no matter where we place the CSV file we will still get the 
```Error: Value : (codeflare-operator.v0.0.0-dev) csv.Spec.Icon elements should contain both data and mediatype``` error when running `make bundle`.

Comment (astefanutti) [reply config/manifests/kustomization.yaml:6]: Hum, that sounds like the CSV hasn't been taken into account. The Icon field is in there no?

Anyhow, the CSV cannot be included into the set of manifests that the ODH operator manages for the codeflare component.

Comment (astefanutti) [reply config/manifests/kustomization.yaml:6]: Thanks, that confirms the CSV wasn't properly taken into account while moved to another directory. We should look into it, so it can be moved out the `manifests` directory, without breaking the OLM bundle generation.

Comment (Bobbins228) [reply config/manifests/kustomization.yaml]: I read up on the `operator-sdk generate bundle` command and I found this.
>A CSV manifest is generated by collecting data from the set of manifests passed to this command (see below),
such as CRDs, RBAC, etc., and applying that data to a "base" CSV manifest. This base CSV can contain metadata,
added by hand or by the 'generate kustomize manifests' command, and can be passed in like any other manifest
(see below) or by file at the exact path '<kustomize-dir>/bases/<package-name>.clusterserviceversion.yaml'.
Be aware that 'generate bundle' idempotently regenerates a bundle, so all non-metadata values in a base
will be overwritten. If no base was passed in, input manifest data will be applied to an empty CSV.

I assume the base in this case would be the CSV file we would normally pass into the manifests folder. So the input data is applied to an empty CSV in our case because we are not supplying. 

I inspected the CSV generated in the bundle folder and the icon and media type are empty strings.

The bundle generates just fine it is just the validation that is an issue.
It seems we can skip certain validation as seen [here](https://github.com/operator-framework/operator-sdk/issues/4375#issuecomment-756896868) 

Doing some more reading on this and I can see there is this documentation on it:
```
To validate a bundle against the validator for operatorhub.io specifically, in addition to required bundle validators:

  $ operator-sdk bundle validate ./bundle --select-optional name=operatorhub
```
If we opted to use this method of validation we would miss out on validating against the operator framework.
What are your thoughts @astefanutti @sutaakar

Comment (sutaakar) [reply config/manifests/kustomization.yaml]: @astefanutti What about keeping CSV in manifests directory and moving ODH related manifests to some other directory?

Comment (astefanutti) [reply config/manifests/kustomization.yaml]: @sutaakar that could be a solution, though that change will have to be coordinated with the ODH operator so it sources the new directory. Actually, it'd be less confusing to have an `odh` or `opendatahub` directory to host these resources, rather than inside the existing `manifests` directory. We need to check with @zdtsw if that's an option.

Comment (zdtsw) [reply config/manifests/kustomization.yaml]: if i understand it correct, what you try to achieve is to do the logic change in your upstream "codeflare-project" and sync down to poendatahub's codeflare-operator with auto update this line  `odh-codeflare-operator-controller-image=quay.io/project-codeflare/codeflare-operator:$somenew_version` ?  then what has it to do with CSV file : thinking : in ODH


in ODH, operator goes to the folder you defined (in current code base is your `config/manifests`) and apply result of  `kustomize build`. I do not think it will matter if you want to restructure your config as long as ODH operator has the correct path with all the resourced defined within.

Comment (astefanutti) [reply config/manifests/kustomization.yaml]: @zdtsw your understanding is correct. The CSV resource must not be part of the directory sourced by ODH. The issue being that operator-sdk is opinionated with using the `manifests` directory, which is currently used by ODH. We would need to change the path for the CodeFlare operator in ODH to something different. Would you have any objections?

Comment (zdtsw) [reply config/manifests/kustomization.yaml]: no objection from my side. 
all we need is when you are ready to have such change in ODH codeflare, let us (operator) know so we can do any modification if needed.

Comment (astefanutti) [reply config/manifests/kustomization.yaml]: Sounds good, thanks @zdtsw!

Comment (Bobbins228) [reply config/manifests/kustomization.yaml]: So I'll put the CSV file back into manifests and should we remove the params files and place them into a new directory for just the ODH CFO?

Comment (astefanutti) [reply config/manifests/kustomization.yaml]: @Bobbins228 yes, and make sure the `odh-codeflare-operator-controller-image` variable is correctly taken into account in the Deployment.

Comment (Bobbins228) [reply config/manifests/kustomization.yaml]: Okay. `manager.yaml` is seemingly the only file that takes in the `IMG` variable into account so what I can do is replace the [image replacement](https://github.com/project-codeflare/codeflare-operator/blob/main/Makefile#L287) with a command that will replace the image in params.env with the `IMG` variable.

I can leave the param files in manager.yaml and it will be updated with the new image.

File: .github/workflows/tag-and-build.yml [hunk 9]
```diff
@@ -64,7 +64,7 @@ jobs:
     - name: Set Go
       uses: actions/setup-go@v3
       with:
-        go-version: v1.20
+        go-version: v1.19
```
sutaakar, line 67: Seems like unintentional change due to rebase.

.github/workflows/tag-and-build.yml
```diff
@@ -102,6 +102,11 @@ jobs:
         sed -i -E "s/(.*INSTASCALE_VERSION \?= ).*/\1${{ github.event.inputs.instascale-version }}/" Makefile
         sed -i -E "s/(.*KUBERAY_VERSION \?= ).*/\1${{ github.event.inputs.kuberay-version }}/" Makefile
 
+    - name: Update image version in params.env
+      run: |
+        VERSION=${{ github.event.inputs.version }} perl -i -pe 's/:(.*)$/:$ENV{"VERSION"}/' ../config/manager/params.env
```
sutaakar, line 107:
```suggestion
        VERSION=${{ github.event.inputs.version }} perl -i -pe 's/:(.*)$/:$ENV{"VERSION"}/' config/manager/params.env
```

Makefile
```diff
@@ -198,7 +198,7 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
-	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	cd config/manager && IMAGE=$(IMG) perl -i -pe 's/odh-codeflare-operator-controller-image=(.*)$$/odh-codeflare-operator-controller-image=$$ENV{"IMAGE"}/' params.env
```
sutaakar, line 201: Just an idea for optional discussion:
As `odh-codeflare-operator-controller-image` parameter is used in upstream, would it have sense to rename it to `codeflare-operator-controller-image` - as it is not ODH specific any more?

config/manager/kustomization.yaml
```diff
@@ -3,3 +3,19 @@ resources:
 
 generatorOptions:
   disableNameSuffixHash: true
+
+configMapGenerator:
+- name: codeflare-stack-config
```
sutaakar, line 8: I think that `stack-config` could be enough.
All resources are automatically prefixed with `codeflare-operator-` - https://github.com/project-codeflare/codeflare-operator/blob/main/config/default/kustomization.yaml#L9

config/manager/kustomization.yaml
```diff
@@ -3,3 +3,19 @@ resources:
...
+  envs:
+    - params.env
+configurations:
+  - params.yaml
+
+vars:
+- name: odh_codeflare_operator_controller_image
+  objref:
+    kind: ConfigMap
+    name: codeflare-stack-config
```
sutaakar, line 18: Related to above.

Makefile
```diff
@@ -283,6 +283,7 @@ validate-bundle: install-operator-sdk
 
 .PHONY: bundle
 bundle: manifests kustomize install-operator-sdk ## Generate bundle manifests and metadata, then validate generated files.
+	cd config/manager && IMAGE=$(IMG) perl -i -pe 's/codeflare-operator-controller-image=(.*)$$/codeflare-operator-controller-image=$$ENV{"IMAGE"}/' params.env
 	$(OPERATOR_SDK) generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
```
sutaakar, line 288:
```suggestion
```

config/manifests/codeflare-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,71 @@
+apiVersion: operators.coreos.com/v1alpha1
```
sutaakar, line 1: ClusterServiceVersion file is now in two places, config/manifests/codeflare-operator.clusterserviceversion.yaml and config/manifests/bases/codeflare-operator.clusterserviceversion.yaml

config/manager/manager.yaml
```diff
@@ -34,7 +34,7 @@ spec:
       containers:
       - command:
         - /manager
-        image: controller:latest
+        image: $(codeflare_operator_controller_image)
```
astefanutti, line 37: Shouldn't it be `codeflare-operator-controller-image` instead of `codeflare_operator_controller_image`?

Makefile
```diff
@@ -198,7 +198,7 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
-	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	cd config/manager && IMAGE=$(IMG) perl -i -pe 's/codeflare-operator-controller-image=(.*)$$/codeflare-operator-controller-image=$$ENV{"IMAGE"}/' params.env
```
astefanutti, line 201: Is that change really needed?

config/manager/params.env
```diff
@@ -0,0 +1 @@
+codeflare-operator-controller-image=quay.io/opendatahub/codeflare-operator:v1.0.1
```
astefanutti, line 1: Do we have an agreement with the ODH operator team to change the name from `odh-codeflare-operator-controller-image`?

config/manager/manager.yaml
Bobbins228: The `codeflare_operator_controller_image` is just a variable for the image in the params file.
We specify the replacement for `$(codeflare_operator_controller_image)` in the [kustomization](https://github.com/project-codeflare/codeflare-operator/pull/398/files#diff-14effe9ea1a3ddb6e419e627a3501a6c42aa2dcbc76803923e9aff2438e6ee34R15) file

Makefile
Bobbins228: Yes we are using the command to update the params file so that the image variable `codeflare-operator-controller-image` is updated with the passed `IMG`

config/manager/params.env
Bobbins228: Good catch I haven't informed anyone yet of this change. Will I just drop a message into the ODH DW channel?

zdtsw: late for the party, but, if i understand this correctly, we are not going to keep CFO in the near future for ODH and RHOAI. 
so the changes done in upstream(as project-codeflare) is irrelevant to ODH and RHOAI.
which means, purely from project-codeflare perspective, these changes are not needed?

PR #406, opened 2023-11-14
Prase command line flags

ronensc:
# Issue link

# What changes have been made

- Added a call to `flag.Parse()` to parse command line flags.
- Updated `glog` module to fix panic on help message (`./bin/manager --help`).

# Verification steps

Running `bin/manager --help` should output:
```
Usage of bin/manager:
  -alsologtostderr
     log to standard error as well as files
  -config string
     The name of the ConfigMap to load the operator configuration from. If it does not exist, the operator will create and initialise it. (default "codeflare-operator-config")
  -kubeconfig string
     Paths to a kubeconfig. Only required if out-of-cluster.
  -log_backtrace_at value
     when logging hits line file:N, emit a stack trace
  -log_dir string
     If non-empty, write log files in this directory
  -log_link string
     If non-empty, add symbolic links in this directory to the log files
  -logbuflevel int
     Buffer log messages logged at this level or lower (-1 means don't buffer; 0 means buffer INFO only; ...). Has limited applicability on non-prod platforms.
  -logtostderr
     log to standard error instead of files
  -stderrthreshold value
     logs at or above this threshold go to stderr (default 2)
  -v value
     log level for V logs
  -vmodule value
     comma-separated list of pattern=N settings for file-filtered logging
```

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/main.go
+++ b/main.go
@@ -81,6 +81,7 @@ func main() {
 	flag.StringVar(&configMapName, "config", "codeflare-operator-config",
 		"The name of the ConfigMap to load the operator configuration from. "+
 			"If it does not exist, the operator will create and initialise it.")
+	flag.Parse()
 
 	zapOptions := zap.Options{
 		Development: true,
--- a/main.go
+++ b/main.go
@@ -81,13 +81,13 @@ func main() {
 	flag.StringVar(&configMapName, "config", "codeflare-operator-config",
 		"The name of the ConfigMap to load the operator configuration from. "+
 			"If it does not exist, the operator will create and initialise it.")
-	flag.Parse()
 
 	zapOptions := zap.Options{
 		Development: true,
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -81,6 +81,7 @@ func main() {
 	flag.StringVar(&configMapName, "config", "codeflare-operator-config",
 		"The name of the ConfigMap to load the operator configuration from. "+
 			"If it does not exist, the operator will create and initialise it.")
+	flag.Parse()
```
astefanutti, line 84: Shouldn't that be called after all the flags have been bound, i.e., after `zapOptions.BindFlags(flag.CommandLine)`?

ronensc: Of course! good catch! thanks
Added a fix

astefanutti:
Thanks!

PR #453, opened 2024-01-25
add raycluster controller to CFO

KPostOffice:
# Issue link
https://issues.redhat.com/browse/RHOAIENG-1992

# What changes have been made
Added a controller to CFO operator which does a series of ServerSideApplies in order to create the requisite k8 resources for securing the RayDashboard

# Verification steps
This PR does not implement the webhook required for automatically injecting the oauth sidecar into the RayCluster definition so that must be done by the user at this point.

TODO Provide a working RayCluster definition for validating the authenticated dashboard

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


KPostOffice:
I tested and made sure that all the expected resources are created. Still need to write unit tests and add a RayCluster def which with a OAuth sidecar that will work with the created resources

KPostOffice:
```yaml
apiVersion: ray.io/v1
kind: RayCluster
metadata:
  labels:
    controller-tools.k8s.io: '1.0'
  annotations:
    codeflare.dev/oauth: 'true'
  name: raytest
  namespace: default
spec:
  enableInTreeAutoscaling: false
  headGroupSpec:
    rayStartParams:
      block: 'true'
      dashboard-host: 0.0.0.0
      num-gpus: '0'
    serviceType: ClusterIP
    template:
      spec:
        containers:
        - env:
          - name: MY_POD_IP
            valueFrom:
              fieldRef:
                fieldPath: status.podIP
          image: quay.io/project-codeflare/ray:latest-py39-cu118
          imagePullPolicy: Always
          lifecycle:
            preStop:
              exec:
                command:
                - /bin/sh
                - -c
                - ray stop
          name: ray-head
          ports:
          - containerPort: 6379
            name: gcs
          - containerPort: 8265
            name: dashboard
          - containerPort: 10001
            name: client
          resources:
            limits:
              cpu: 1
              memory: 8G
              nvidia.com/gpu: 0
            requests:
              cpu: 1
              memory: 8G
              nvidia.com/gpu: 0
        - args:
          - --https-address=:8443
          - --provider=openshift
          - --openshift-service-account=raytest
          - --upstream=http://localhost:8265
          - --tls-cert=/etc/tls/private/tls.crt
          - --tls-key=/etc/tls/private/tls.key
          - --cookie-secret=pijwpRyx9a2FaLg3PqtvrSFO9WeItVFe0vOYYLR5RbcZCdR04aKjROd/kUrtqCtJnRlGeXh0MW0vSbVvWvCjDg==
          - --openshift-delegate-urls={"/":{"resource":"pods","namespace":"default","verb":"get"}}
          image: registry.redhat.io/openshift4/ose-oauth-proxy@sha256:1ea6a01bf3e63cdcf125c6064cbd4a4a270deaf0f157b3eabb78f60556840366
          name: oauth-proxy
          ports:
          - containerPort: 8443
            name: oauth-proxy
          resources: {}
          volumeMounts:
          - mountPath: /etc/tls/private
            name: proxy-tls-secret
            readOnly: true
        imagePullSecrets: []
        serviceAccount: raytest
        volumes:
        - name: proxy-tls-secret
          secret:
            secretName: raytest-tls
  rayVersion: 2.7.0
  workerGroupSpecs:
  - groupName: small-group-raytest
    maxReplicas: 1
    minReplicas: 1
    rayStartParams:
      block: 'true'
      num-gpus: '0'
    replicas: 1
    template:
      metadata:
        annotations:
          key: value
        labels:
          key: value
      spec:
        containers:
        - env:
          - name: MY_POD_IP
            valueFrom:
              fieldRef:
                fieldPath: status.podIP
          image: quay.io/project-codeflare/ray:latest-py39-cu118
          lifecycle:
            preStop:
              exec:
                command:
                - /bin/sh
                - -c
                - ray stop
          name: machine-learning
          resources:
            limits:
              cpu: 1
              memory: 4G
              nvidia.coma/gpu: 0
            requests:
              cpu: 1
              memory: 4G
              nvidia.com/gpu: 0
        imagePullSecrets: []
```
Here is a well formatted RayCluster that can be used to test these changes. Run the controller using:
```
$ NAMESPACE=foobar make run
```
Then apply the yaml above. There should be a route in the same namespace on the cluster for the dashboard that is protected by an OAuth Proxy

KPostOffice:
/hold

Still need to edit CI/makefile so that the new Envtest tests will run

config/rbac/role.yaml
```diff
@@ -5,219 +5,62 @@ metadata:
   creationTimestamp: null
```
astefanutti, line 5: I don't think generation of that file work correctly at the moment. RBAC for MCAD has to be declared for it to work.

config/crd/test/ray.io_rayclusters.yaml
```diff
@@ -0,0 +1,14234 @@
+---
```
astefanutti, line 1: Can we use a reference to import these CRDs as we do for MCAD: https://github.com/project-codeflare/codeflare-operator/blob/1264faabc1835b3aafb5e11fc31c4b1bbb9a1382/config/crd/mcad/kustomization.yaml#L4C6-L4C83

controllers/raycluster_controller.go
```diff
@@ -0,0 +1,325 @@
...
+		WithOwnerReferences(
+			v1.OwnerReference().WithUID(cluster.UID).WithName(cluster.Name).WithKind(cluster.Kind).WithAPIVersion(cluster.APIVersion),
+		)
+}
+
+func desiredClusterRoute(cluster *rayv1.RayCluster) *routeapply.RouteApplyConfiguration {
+	return routeapply.Route(cluster.Name, cluster.Namespace).
+		WithLabels(map[string]string{"ray.io/cluster-name": cluster.Name}).
+		WithSpec(routeapply.RouteSpec().
+			WithTo(routeapply.RouteTargetReference().WithKind("Service").WithName(cluster.Name + "tls")).
```
astefanutti, line 254: Where is that service created? I'd advocate to have it created by this controller as well.

controllers/raycluster_controller.go
```diff
@@ -0,0 +1,325 @@
...
+			ctx, types.NamespacedName{Namespace: cluster.Namespace, Name: cluster.Name}, &routev1.Route{},
+		)
+		if err != nil {
+			logger.Error(err, "Error deleting OAuth Route, retrying", logRequeueing, strTrue)
+			return ctrl.Result{Requeue: true, RequeueAfter: requeueTime}, nil
+		}
+		return ctrl.Result{}, nil
+	}
+
+	routeClient, err := routev1client.NewForConfig(r.k8Config)
```
astefanutti, line 179: Have that client as a member of the reconciler struct.

astefanutti (commented):
Two points:
* Do we want to leverage static sidecar injection in the Kuberay deployment?
* Could we have a e2e test that run on OpenShift?

@sutaakar FYI.

controllers/raycluster_controller.go
KPostOffice: https://github.com/project-codeflare/codeflare-operator/pull/453/files#diff-9514176087ba6764592f4797d22a2ef7e9c4b0204ecfc94efa00d371b66f6bb7R271

It's already created by this controller

config/crd/test/ray.io_rayclusters.yaml
KPostOffice: We only need it for testing and I couldn't get envtest to work with remote files. I'm thinking of changing it so that it downloads them and cleans them up as part of the `BeforeSuite` and `AfterSuite`. It really only needs the `route` and the `raycluster`. Do we need to have the Ray CRDs applied as well before starting the controller?

config/rbac/role.yaml
KPostOffice: Good catch. I may just remake this by hand

astefanutti: Yes, manual edit is fine at the moment.

config/crd/test/ray.io_rayclusters.yaml
astefanutti: Ah I missed that envtest cannot use Kustomize reference mechanism.

For the dependency on the RayCluster API and the KubeRay CRD, I have some ideas that I need to formalise, but in the short term, what we can do is only start the controller when the API is present in the cluster.

controllers/raycluster_controller.go
astefanutti: Ah right, got confused somehow :).

main.go
```diff
@@ -170,6 +177,8 @@ func main() {
 		}
 		exitOnError(instaScaleController.SetupWithManager(context.Background(), mgr), "Error setting up InstaScale controller")
 	}
+	rayClusterController := cfoControllers.RayClusterReconciler{Client: mgr.GetClient(), Scheme: mgr.GetScheme()}
```
astefanutti, line 180: Let's add the controller conditionally to the presence of the RayCluster API on the cluster?

controllers/raycluster_controller_test.go
```diff
@@ -0,0 +1,222 @@
...
+
+func randSeq(n int) string {
+	b := make([]rune, n)
+	for i := range b {
+		b[i] = letters[r.Intn(len(letters))]
+	}
+	return string(b)
+}
+
+var _ = Describe("RayCluster controller", func() {
```
sutaakar, line 57: A personal comment - do you have a specific reason or preference to use Ginkgo?
I personally find it less readable than "usual" tests, but that may be just my own opinion.

sutaakar:
just one note, when I delete RayCluster the controller logs error message:
```
2024-02-08T15:59:15+01:00	ERROR	Cluster does not exist	{"controller": "codeflare-raycluster-controller", "controllerGroup": "ray.io", "controllerKind": "RayCluster", "RayCluster": {"name":"raytest","namespace":"ksuta-test"}, "namespace": "ksuta-test", "name": "raytest", "reconcileID": "a5481a21-a1b5-4ced-a7b9-0aa3b355bc3f", "error": "RayCluster.ray.io \"raytest\" not found"}
github.com/project-codeflare/codeflare-operator/controllers.(*RayClusterReconciler).Reconcile
	/home/ksuta/src/github.com/project-codeflare/codeflare-operator/controllers/raycluster_controller.go:109
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
	/home/ksuta/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.3/pkg/internal/controller/controller.go:118
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/home/ksuta/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.3/pkg/internal/controller/controller.go:314
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/home/ksuta/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.3/pkg/internal/controller/controller.go:265
```
It may be good to skip it or skip events which are not relevant.

controllers/raycluster_controller_test.go
KPostOffice: Ginkgo is the framework included by default by the OperatorSDK framework. I figured Ginkgo would be best practice

main.go
KPostOffice: On second thought, can you address my comment above?:
> This can cause race conditions when both components are enabled. If CFO starts before KubeRay, then this Reconciler will never start despite the API being present soon after

config/crd/test/ray.io_rayclusters.yaml
astefanutti: Right, but on the other hand, the CFO will never work if KubeRay is not installed. I don't think there is an ideal solution to this problem in the short term. We can document the CFO has to be restarted after KubeRay is installed, until we rework the deployment strategy.

My current thinking is that this controller should be deployed along side KubeRay, e.g. in a sidecar container.

controllers/raycluster_controller_test.go
sutaakar: ok

sutaakar:
tests seem to need some quality of life adjustments, will improve them

ChristianZaccaria (changes requested):

controllers/raycluster_controller.go
```diff
@@ -0,0 +1,321 @@
...
+}
+
+const (
+	requeueTime             = 10
+	controllerName          = "codeflare-raycluster-controller"
+	oauthAnnotation         = "codeflare.dev/oauth=true"
+	CodeflareOAuthFinalizer = "codeflare.dev/oauth-finalizer"
+	OAuthServicePort        = 443
+	OAuthServicePortName    = "oauth-proxy"
+	OAuthProxyImage         = "registry.redhat.io/openshift4/ose-oauth-proxy:latest"
```
ChristianZaccaria, line 63: As previously discussed, this can be removed

ChristianZaccaria (changes requested):

KPostOffice:
@sutaakar I'm not sure what is causing the `OLM Install and Upgrade` and the `e2e` checkruns to fail. I don't think I've really touched anything that should effect either, were they broken before this PR or did something in this PR break them? The verify generated files is expected to fail since the generated role doesn't contain all necessary permissions for the controller.

main.go
KPostOffice: @astefanutti Can you take a look at the conditional I've added for this? I was having trouble finding other examples and I'm not sure the way I did it is best.

sutaakar:
@KPostOffice e2e tests are using generated role, so failed generated role is affecting tests.

main.go
```diff
@@ -170,6 +182,18 @@ func main() {
 		}
 		exitOnError(instaScaleController.SetupWithManager(context.Background(), mgr), "Error setting up InstaScale controller")
 	}
+	groups, err := kubeClient.DiscoveryClient.ServerGroups()
+	exitOnError(err, "Could not get server groups and resources")
+	// fmt.Println(resources)
+	for _, g := range groups.Groups {
```
astefanutti, line 188: I'd suggest to create a separate util function to encapsulate the logic, e.g., something like:

```go
func HasAPIResourceForGVK(dc discovery.DiscoveryInterface, gvk schema.GroupVersionKind) (bool, error) {
	gv, kind := gvk.ToAPIVersionAndKind()
	if resources, err := dc.ServerResourcesForGroupVersion(gv); err != nil {
		if apierrors.IsNotFound(err) {
			return false, nil
		}
		return false, err
	} else {
		for _, res := range resources.APIResources {
			if res.Kind == kind {
				return true, nil
			}
		}
	}
	return false, nil
}
```

KPostOffice: Thanks @astefanutti

KPostOffice:
> @KPostOffice e2e tests are regenerating role, so failed generated role is affecting tests.

I've added a manually generated ClusterRole file with the old contents that way the generated files succeed

sutaakar:
@KPostOffice Can you also create ClusterRoleBinding which will connect ServiceAccount with the new role?
Something like https://github.com/project-codeflare/codeflare-operator/blob/main/config/rbac/role_binding.yaml

main.go
```diff
@@ -67,6 +72,11 @@ var (
 	BuildDate         = "UNKNOWN"
 )
 
+const (
+	RayGroup       = "ray.io"
+	RayClusterKind = "RayCluster"
+)
```
astefanutti, line 78: Is it actually used?

controllers/raycluster_controller.go
```diff
@@ -0,0 +1,320 @@
...
+		WithOwnerReferences(
+			v1.OwnerReference().WithUID(cluster.UID).WithName(cluster.Name).WithKind(cluster.Kind).WithAPIVersion(cluster.APIVersion),
+		)
+}
+
+func oauthSecretNameFromCluster(cluster *rayv1.RayCluster) string {
+	return cluster.Name + "-oauth-config"
+}
+
+// NewClusterOAuthSecret defines the desired OAuth secret object
```
astefanutti, line 290: `NewClusterOAuthSecret` -> `desiredOAuthSecret`

config/rbac/manual_role.yaml
```diff
@@ -0,0 +1,223 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  name: manual-manager-role
```
astefanutti, line 6: The name is a bit confusing. What about skipping role generation altogether at the moment, and keep the original role manifest with the extra rules added manually?

controllers/raycluster_controller.go
```diff
@@ -0,0 +1,320 @@
...
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	if cluster.ObjectMeta.DeletionTimestamp.IsZero() {
+		if !controllerutil.ContainsFinalizer(&cluster, CodeflareOAuthFinalizer) {
+			logger.Info("Add a finalizer", "finalizer", CodeflareOAuthFinalizer)
+			controllerutil.AddFinalizer(&cluster, CodeflareOAuthFinalizer)
+			if err := r.Update(ctx, &cluster); err != nil {
+				logger.Error(err, "Failed to update RayCluster with finalizer", logRequeueing, strTrue)
+				return ctrl.Result{Requeue: true, RequeueAfter: requeueTime}, err
```
ChristianZaccaria, line 109: It seems based on documentation that the `Requeue` bool is not necessary to be set as it is implied when using `RequeueAfter`. It could be removed on all instances.

https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.15.3/pkg/reconcile#Result

ChristianZaccaria (changes requested):

main.go
KPostOffice: Ahh, nope. Not anymore

config/rbac/manual_role.yaml
KPostOffice: That's what I did at first, but it's failing CI since it checks against generated files and also regenerates files before running tests.

astefanutti: Ah right, it wasn't generating anything before because there were no kubebuilder markers.

What about naming it `mead-manager-role`? I think that's more descriptive.

controllers/raycluster_controller.go
```diff
@@ -0,0 +1,320 @@
...
+				WithKind("ClusterRole").
+				WithName("system:auth-delegator"),
+		).
+		WithOwnerReferences(
+			v1.OwnerReference().WithUID(cluster.UID).WithName(cluster.Name).WithKind(cluster.Kind).WithAPIVersion(cluster.APIVersion),
+		)
+}
+
+func desiredServiceAccount(cluster *rayv1.RayCluster) *coreapply.ServiceAccountApplyConfiguration {
+	return coreapply.ServiceAccount(cluster.Name, cluster.Namespace).
```
ChristianZaccaria, line 233: Just to make sure, is it expected for the service account name and other resources to differ in the values from the SDK? I.e., the name here would be `cluster.Name`, while in the SDK we would have [`cluster.Name-oauth-proxy`](https://github.com/project-codeflare/codeflare-sdk/blob/main/src/codeflare_sdk/utils/generate_yaml.py#L557).

Bobbins228 (changes requested):

controllers/raycluster_controller.go
```diff
@@ -0,0 +1,320 @@
...
+				`{"kind":"OAuthRedirectReference","apiVersion":"v1",` +
+				`"reference":{"kind":"Route","name":"` + cluster.Name + `"}}`,
+		}).
+		WithOwnerReferences(
+			v1.OwnerReference().WithUID(cluster.UID).WithName(cluster.Name).WithKind(cluster.Kind).WithAPIVersion(cluster.APIVersion),
+		)
+}
+
+func desiredClusterRoute(cluster *rayv1.RayCluster) *routeapply.RouteApplyConfiguration {
+	return routeapply.Route(cluster.Name, cluster.Namespace).
```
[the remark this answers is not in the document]
KPostOffice: ++. Note though, I think the SDK should get the value directly from the Route. i.e. find route with label `ray.io/cluster-name: cluster.Name`

controllers/raycluster_controller.go
KPostOffice: this shouldn't really matter since it is just internals and won't create breaking changes for the user but I will adjust them to be consistent

Bobbins228 (commented):
/lgtm

sutaakar (commented):
/lgtm

ChristianZaccaria (changes requested):

controllers/raycluster_controller.go
```diff
@@ -0,0 +1,327 @@
...
+		controllerutil.RemoveFinalizer(&cluster, CodeflareOAuthFinalizer)
+		if err := r.Update(ctx, &cluster); err != nil {
+			logger.Error(err, "Failed to remove finalizer from RayCluster", logRequeueing, strTrue)
+			return ctrl.Result{RequeueAfter: requeueTime}, err
+		}
+		logger.Info("Successfully removed finalizer.", logRequeueing, strFalse)
+		return ctrl.Result{}, nil
+	}
+
+	if val, ok := cluster.ObjectMeta.Annotations["codeflare.dev/oauth"]; !ok || val != "true" {
```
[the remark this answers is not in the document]
KPostOffice: I'll turn the string to lowercase here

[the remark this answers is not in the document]
ChristianZaccaria: To uppercase would be best as the annotation in the RayCluster when created ends up being `codeflare.dev/oauth: 'True'`

controllers/raycluster_controller.go
```diff
@@ -0,0 +1,327 @@
...
+				WithTermination(routev1.TLSTerminationReencrypt),
+			),
+		).
+		WithOwnerReferences(
+			v1.OwnerReference().WithUID(cluster.UID).WithName(cluster.Name).WithKind(cluster.Kind).WithAPIVersion(cluster.APIVersion),
+		)
+}
+
+func oauthServiceNameFromCluster(cluster *rayv1.RayCluster) string {
+	return cluster.Name + "-oauth"
```
ChristianZaccaria, line 270:
```suggestion
	return cluster.Name + "-proxy-tls-secret"
```

ChristianZaccaria (changes requested):

controllers/raycluster_controller.go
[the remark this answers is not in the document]
KPostOffice: Sorry. I meant for checking i.e. val.Lower = "true"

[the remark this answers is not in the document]
KPostOffice: I ended up finding this: https://pkg.go.dev/strconv#ParseBool which will be better in the long run since it allows for any truthy value of the annotation to be interpreted correctly. If the value isn't interpretable by the package  (i.e. returns an error) the default behavior is to treat it as false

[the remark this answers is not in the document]
ChristianZaccaria: That is actually pretty cool, nice one Kevin!

Bobbins228 (commented):
Tested this with Christians work on the [SDK](https://github.com/project-codeflare/codeflare-sdk/pull/480) works as expected

/lgtm

KPostOffice:
I changed adding the taint to be `info` level in the case of an error since errors are not unexpected due to the two controllers managing the resource

PROJECT
```diff
@@ -6,4 +10,10 @@ plugins:
   scorecard.sdk.operatorframework.io/v2: {}
 projectName: codeflare-operator
 repo: github.com/project-codeflare/codeflare-operator
+resources:
+- controller: true
+  domain: codeflare.dev
```
astefanutti, line 15: Maybe a nit, but I'm not sure that's the right domain here if it's needed.

controllers/raycluster_controller.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+func (r *RayClusterReconciler) deleteIfNotExist(ctx context.Context, namespacedName types.NamespacedName, obj client.Object) error {
+	err := r.Client.Get(ctx, namespacedName, obj)
+	if err != nil && errors.IsNotFound(err) {
+		return nil
+	} else if err != nil {
+		return err
+	}
+	return r.Client.Delete(ctx, obj, &deleteOptions)
```
astefanutti, line 169: Isn't it simpler to ignore not found error on delete?

pkg/config/config.go
```diff
@@ -36,6 +36,8 @@ type CodeFlareOperatorConfiguration struct {
 
 	// The InstaScale controller configuration
 	InstaScale *InstaScaleConfiguration `json:"instascale,omitempty"`
+
+	RayClusterOAuth *bool `json:"rayClusterOAuth,omitempty"`
```
astefanutti, line 40: I'd suggest  to structure the configuration a bit, in a dedicated struct.

controllers/raycluster_controller.go
```diff
@@ -0,0 +1,292 @@
...
+
+const (
+	requeueTime             = 10
+	controllerName          = "codeflare-raycluster-controller"
+	oauthAnnotation         = "codeflare.dev/oauth=true"
+	CodeflareOAuthFinalizer = "codeflare.dev/oauth-finalizer"
+	OAuthServicePort        = 443
+	OAuthServicePortName    = "oauth-proxy"
+	strTrue                 = "true"
+	strFalse                = "false"
```
astefanutti, line 64: That can be removed?

controllers/raycluster_controller.go
```diff
@@ -0,0 +1,292 @@
...
+	kubeClient  *kubernetes.Clientset
+	routeClient *routev1client.RouteV1Client
+	Scheme      *runtime.Scheme
+	CookieSalt  string
+}
+
+const (
+	requeueTime             = 10
+	controllerName          = "codeflare-raycluster-controller"
+	oauthAnnotation         = "codeflare.dev/oauth=true"
```
astefanutti, line 59: That can be removed?

controllers/raycluster_controller.go
```diff
@@ -0,0 +1,292 @@
...
+	CodeflareOAuthFinalizer = "codeflare.dev/oauth-finalizer"
```
astefanutti, line 60: May I suggest something like `openshift.ai/oauth-proxy`? In case we move the controller outside codeflare.

pkg/config/config.go
```diff
@@ -37,6 +37,10 @@ type CodeFlareOperatorConfiguration struct {
 	// The InstaScale controller configuration
 	InstaScale *InstaScaleConfiguration `json:"instascale,omitempty"`
 
+	Authorization *AuthorizationConfiguration `json:"authorization,omitempty"`
```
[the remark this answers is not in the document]
KPostOffice: Ahh, got it. Will fix

pkg/config/config.go
```diff
@@ -36,6 +36,12 @@ type CodeFlareOperatorConfiguration struct {
 
 	// The InstaScale controller configuration
 	InstaScale *InstaScaleConfiguration `json:"instascale,omitempty"`
+
+	KubeRay *KubeRayConfiguration `json:"authorization,omitempty"`
```
astefanutti, line 40: ``json:"authorization,omitempty"`` -> ``json:"kuberay,omitempty"``

pkg/config/config.go
```diff
@@ -36,6 +36,12 @@ type CodeFlareOperatorConfiguration struct {
...
+}
+
+type KubeRayConfiguration struct {
+	RayDashboardOAuthEnabled *bool `json:"rayClusterOAuth,omitempty"`
```
astefanutti, line 44: ``json:"rayClusterOAuth,omitempty"`` -> ``json:"rayClusterOAuthEnabled,omitempty"``

astefanutti:
LGTM, we may want to be more fine-grained to enable the creation of Route without OAuth, but that can be done later.

PR #455, opened 2024-01-26
Instascale E2E test for nodepools

Fiona-Waters:
# Issue link
[Jira Issue](https://issues.redhat.com/browse/RHOAIENG-1312)

# What changes have been made
An e2e test has been added to test the entire instascale flow on a ROSA Hosted Hypershift Cluster.

I have also included an update to the `instaScaleJobAppWrapper` function based on recent changes to the mnist.py script and other e2e tests.

# Verification steps
 - Provision a [Hypershift Cluster](https://docs.google.com/document/d/1x1GM7NoR5x53ESTZ-nKZa2o8MOmBVSWo5ndR0oVKPuI/edit#heading=h.aqaecxvsy88r)
 -  Install RHODS via Operator Hub setting the CodeFlare stack to `Managed`
 - Create a separate Instascale deployment using the current main branch to include recent updates regarding naming convention. 
     - Make sure to add the `instascale-ocm-secret` in the default namespace.
 -  Install NFD operator with NodeFeatureDiscovery CR.
 - Install NVIDIA operator with ClusterPolicy CR.
 - Run/debug the test and confirm that it passes.

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [X] Manual tests

### DIFF
--- a/test/e2e/instascale_nodepool_test.go
+++ b/test/e2e/instascale_nodepool_test.go
@@ -1,0 +1,66 @@
+package e2e
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+	. "github.com/project-codeflare/codeflare-common/support"
+	mcadv1beta1 "github.com/project-codeflare/multi-cluster-app-dispatcher/pkg/apis/controller/v1beta1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestInstascaleNodepool(t *testing.T) {
+
+	test := With(t)
+	test.T().Parallel()
+
+	clusterType := GetClusterType(test)
+	if clusterType != HypershiftCluster {
+		test.T().Skipf("Skipping test as not running on an Hypershift cluster, resolved cluster type: %s", clusterType)
+	}
+
+	namespace := test.NewTestNamespace()
+
+	// Test configuration
+	cm := CreateConfigMap(test, namespace.Name, map[string][]byte{
+		// pip requirements
+		"requirements.txt": ReadFile(test, "mnist_pip_requirements.txt"),
+		// MNIST training script
+		"mnist.py": ReadFile(test, "mnist.py"),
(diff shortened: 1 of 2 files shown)


test/e2e/instascale_nodepool_test.go
```diff
@@ -0,0 +1,66 @@
...
+	})
+
+	//create OCM connection
+	connection := CreateOCMConnection(test)
+	defer connection.Close()
+
+	// check existing cluster resources
+	// look for a node pool with a label key equal to aw name - expect NOT to find it
+	test.Expect(GetNodePools(test, connection)).
+		ShouldNot(ContainElement(WithTransform(NodePoolLabels, HaveKey(HavePrefix("test-instascale")))))
```
VanillaSpoon, line 40: The label here has been updated to `name-namespace` to prevent conflicts.
we could do something similar to: 
```go
// look for a node pool with a label key equal to aw.name-aw.namespace - expect NOT to find it

expectedLabel = fmt.Sprintf("%s-%s", appWrapperName, appWrapperNamespace)

test.Expect(GetNodePools(test, connection)).
	ShouldNot(ContainElement(WithTransform(NodePoolLabels, HaveKey(expectedLabel))))

```

However, this still works, just thought it worth mentioning :)

VanillaSpoon (commented):
Hey Fiona, 
Just left a small comment for considerations, this would be up to you though :)

Tests run as exected :) \lgtm

test/e2e/instascale_nodepool_test.go
Fiona-Waters: Thanks Eoin! I think using `HavePrefix` is neater but we should use whichever is best practice. What do you think @sutaakar?

Fiona-Waters: Updated, thank you both.

sutaakar (commented):
/lgtm

PR #478, opened 2024-02-20
Display input for KubeRay parameter and check for existence

ChristianZaccaria:
# Issue link

# What changes have been made

Display KubeRay parameter in release workflow and check for the existence of the chosen KubeRay release.

Previously the KubeRay parameter was not displayed
image

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/.github/workflows/project-codeflare-release.yml
+++ b/.github/workflows/project-codeflare-release.yml
@@ -51,12 +51,14 @@ jobs:
         echo "MCAD Version: ${{ github.event.inputs.mcad-version }}"
         echo "CodeFlare SDK Version: ${{ github.event.inputs.codeflare-sdk-version }}"
         echo "InstaScale Version: ${{ github.event.inputs.instascale-version }}"
+        echo "Tested KubeRay Version: ${{ github.event.inputs.kuberay-version }}"
         echo "Is Stable: ${{ github.event.inputs.is-stable }}"
         echo "CodeFlare Repository Organization: ${{ github.event.inputs.codeflare-repository-organization }}"
         echo "Quay Organization: ${{ github.event.inputs.quay-organization }}"
         echo "Community Operators Prod Organization: ${{ github.event.inputs.community-operators-prod-organization }}"
 
   release-mcad:
+    needs: check-kuberay-version
     runs-on: ubuntu-latest
 
     steps:
@@ -120,6 +122,7 @@ jobs:
       if: ${{ env.INSTASCALE_RELEASE_STATUS_CODE != '200' }}
 
   release-codeflare-sdk:
+    needs: check-kuberay-version
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/project-codeflare-release.yml
```diff
@@ -186,3 +187,15 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Check if KubeRay release exists
+      run: |
+        status_code=$(curl -s -o /dev/null -w "%{http_code}" https://github.com/ray-project/kuberay/releases/tag/${{ github.event.inputs.kuberay-version }})
+        if [[ "$status_code" == "200" ]]; then
+          echo "KubeRay release with version ${{ github.event.inputs.mcad-version }} exists and available to use."
+        fi
+        echo "KubeRay release with version ${{ github.event.inputs.kuberay-version }} does not exist. Please select an existing version."
```
sutaakar, line 201: Can you make sure that the step fails in case KubeRay version is not available?

sutaakar:
@ChristianZaccaria Thinking about usage of `Check if KubeRay release exists`.
Should it be just informative or should it be considered as a prerequisite of the release?

ChristianZaccaria:
> @ChristianZaccaria Thinking about usage of `Check if KubeRay release exists`. Should it be just informative or should it be considered as a prerequisite of the release?

I think in this case we will always want to specify an existing version of KubeRay to ensure the compatibility matrix is updated to the correct version we have tested. Inputting a non existing version in the matrix could cause confusion. What do you think?

sutaakar:
In such case it may be good to fail whole pipeline in case KubeRay version is not valid.
Can you adjust other steps to require this new step to pass first before being triggered?

ChristianZaccaria:
I added that check now. I think we don't really require the `check-kuberay-version` to be triggered for instascale or codeflare-operator as the other components will need to pass first. Should I keep the check or remove them from instascale and CFO?

sutaakar:
You can remove it from InstaScale and operator

ChristianZaccaria:
Updated, I think it's good to go. Thanks Karel

.github/workflows/project-codeflare-release.yml
```diff
@@ -186,3 +189,17 @@ jobs:
...
+
+  check-kuberay-version:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Check if KubeRay release exists
+      run: |
+        status_code=$(curl -s -o /dev/null -w "%{http_code}" https://github.com/ray-project/kuberay/releases/tag/${{ github.event.inputs.kuberay-version }})
+        if [[ "$status_code" == "200" ]]; then
+          echo "KubeRay release with version ${{ github.event.inputs.mcad-version }} exists and available to use."
```
sutaakar, line 201:
```suggestion
          echo "KubeRay release with version ${{ github.event.inputs.kuberay-version }} exists and available to use."
```

sutaakar:
One nitpick above,otherwise looks fine.

ChristianZaccaria:
Thanks good catch!

PR #485, opened 2024-03-06
update go version from 1.20 to 1.21

dgrove-oss:
1. Go 1.22 has been released, meaning go 1.20 is out of support.
2. Kueue and AppWrapper both have go 1.21 as their minimal version.

### DIFF
--- a/.github/workflows/project-codeflare-release.yml
+++ b/.github/workflows/project-codeflare-release.yml
@@ -72,7 +72,7 @@ jobs:
 
     - name: Release MCAD
       run: |
-        gh workflow run mcad-release.yml --repo ${{ github.event.inputs.codeflare-repository-organization }}/multi-cluster-app-dispatcher --ref go-1.20 --field tag=${{ github.event.inputs.mcad-version }}
+        gh workflow run mcad-release.yml --repo ${{ github.event.inputs.codeflare-repository-organization }}/multi-cluster-app-dispatcher --ref go-1.21 --field tag=${{ github.event.inputs.mcad-version }}
       env:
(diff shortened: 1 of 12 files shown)


ChristianZaccaria (changes requested):
There are several other instances where we could upgrade go versions too including 
https://github.com/project-codeflare/codeflare-operator/blob/bb2460797387b02d918d79c13360801b7de8434d/Dockerfile#L2 
and others, maybe the workflows go version used too.

.github/workflows/project-codeflare-release.yml
```diff
@@ -72,7 +72,7 @@ jobs:
 
     - name: Release MCAD
       run: |
-        gh workflow run mcad-release.yml --repo ${{ github.event.inputs.codeflare-repository-organization }}/multi-cluster-app-dispatcher --ref go-1.20 --field tag=${{ github.event.inputs.mcad-version }}
+        gh workflow run mcad-release.yml --repo ${{ github.event.inputs.codeflare-repository-organization }}/multi-cluster-app-dispatcher --ref go-1.21 --field tag=${{ github.event.inputs.mcad-version }}
```
ChristianZaccaria, line 75: This can remain go-1.20. That is the branch name, there currently isn't a go-1.21 branch in `multi-cluster-app-dispatcher`

ChristianZaccaria (changes requested):

.github/workflows/project-codeflare-release.yml
dgrove-oss: reverted this change.

dgrove-oss:
Closing this.  We're going to take a less incremental approach of migrating the code flare operator from MCADv1 to the new Kueue-capable AppWrapper.  Will open a new PR when the larger step is ready to review.

PR #491, opened 2024-03-19
Upgrade from AppWrapper v1beta1 to v1beta2

dgrove-oss:
# Issue link

# What changes have been made

Context: This PR supports the migration of the codeflare operator from MCADv1 to Kueue+AppWrapper

Related PRs: 
   + https://github.com/project-codeflare/codeflare-sdk/pull/484
   + https://github.com/project-codeflare/codeflare-common/pull/37

Changes:
1. Upgrade AppWrapper from v1beta1 to v1beta2
2. Remove MCADv1 controller
3. Add AppWrapper controller
4. Remove InstaScale controller
5. Update build, test, and CI accordingly

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


dgrove-oss:
Don't bother reviewing this yet.  It will take a few iterations to get in synch with the SDK PR.

dgrove-oss:
This PR is now ready for review.

The Upgrade test is failing, but assuming I understand what the test is doing, I think that is expected and unavoidable.  We explicitly decided we would not attempt to support upgrading from AppWrapper v1beta1 to v1beta2.   We can't just do an in place patch of the container image because more than that has changed.  There are different CRDs and different RBACs.

dgrove-oss:
PR ready for review.  

The PR effectively disables the `OLM Insall and Upgrade` in a [second commit](https://github.com/project-codeflare/codeflare-operator/pull/491/commits/2d56a930f59101bcef59167d1b097b229e56e0cd) to get us through the transition.  Once we flow the changes all the way through the system and there is a codeflare operator release that doesn't contain MCAD/Instascale controllers, we could adjust and re-enable.

Dockerfile
```diff
@@ -1,5 +1,26 @@
 # Build the manager binary
-FROM registry.access.redhat.com/ubi8/go-toolset:1.20.10 as builder
+#FROM registry.access.redhat.com/ubi8/go-toolset:1.20.10 as builder
+
+# BEGIN -- workaround lack of go-toolset for golang 1.21
+
+# FROM registry-proxy.engineering.redhat.com/rh-osbs/openshift-golang-builder:v1.21 AS golang
+FROM golang:1.21 AS golang
+FROM registry.access.redhat.com/ubi8/ubi:8.8 AS builder
```
KPostOffice, line 8: @dimakis Will this be okay in our build pipelines?

KPostOffice (commented):
Looks good! a few questions

.github/workflows/olm_tests.yaml
```diff
@@ -122,80 +122,83 @@ jobs:
...
-
-      - name: Upload logs
-        uses: actions/upload-artifact@v4
-        if: always() && steps.deploy.outcome == 'success'
-        with:
-          name: logs
-          retention-days: 10
-          path: |
-            ${{ env.CODEFLARE_TEST_OUTPUT_DIR }}/**/*.log
+      # TODO: Disabled during the MCAD to Kueue transition.
```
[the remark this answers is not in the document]
dgrove-oss: We'll need a release of the codeflare operator with the new code.  After that is done, we should be able to re-enable.

.github/workflows/precommit.yml
```diff
@@ -11,28 +11,38 @@ on:
...
+          python-version: '3.9'
+      - name: Install pre-commit
+        run: |
+          pip install pre-commit
+
+      # - name: Activate cache
+      #   uses: actions/cache@v4
+      #   with:
+      #     path: /cache
+      #     key: ${{ runner.os }}-cache-${{ hashFiles('**/go.sum', '.pre-commit-config.yaml') }}
```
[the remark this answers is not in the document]
dgrove-oss: The golint in precommit checks kept insisting on finding a go 1.20 environment and failing.  Disabling the cache was the only hammer I found that seemed to convince it to use the 1.21 toolchain/modules that we installed above.  Maybe there is some other way to get the cache to update, but I didn't find it.  Happy to try if there are suggestions.

config/manager/manager.yaml
```diff
@@ -47,6 +47,9 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
+          # TODO: Disabling webhooks is a temporary hack.  Webhooks are required for Kueue integration.
```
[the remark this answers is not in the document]
dgrove-oss: I was thinking a future one, but I had a productive conversation with @astefanutti earlier today about [cert-controller](https://github.com/open-policy-agent/cert-controller).  As a result, I'm going to try to fix in this PR (ideally within the next 24 hours).

pkg/config/config.go
```diff
@@ -30,25 +29,12 @@ type CodeFlareOperatorConfiguration struct {
 
 	// ControllerManager returns the configurations for controllers
 	ControllerManager `json:",inline"`
```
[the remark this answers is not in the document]
dgrove-oss: Yeah.  Don't know what I was thinking... Fixed.

Dockerfile
dgrove-oss: This was based on a Dockerfile I got from @astefanutti for Kueue, but his file used a redhat internal  image AS `golang` that I didn't have access to.  I used the default `golang:1.21` to get unblocked, but its almost certainly not the right to use in the long run.

dgrove-oss:
So we aren't blocked on merging the PR, I've added a new RayCluster test that bypasses AppWrappers entirely and submits the RayCluster directly to Kueue.

dgrove-oss:
@astefanutti @KPostOffice -- this PR is ready for review again.   I've rebased on of the RayCluster PRs, resolved conflicts,  and the operator appears to be working as expected -- e2e tests pass, and I've done some manual AppWrapper testing that looks good.    As discussed, we can't merge this PR until the SDK PRs are ready too, but we want to get it to be ready to go when that happens.

config/manager/manager.yaml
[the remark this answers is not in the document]
dgrove-oss: This is now done.

Dockerfile
dimakis: no, we won't get away with that. we'll need to go back to a RH registry

.github/workflows/olm_tests.yaml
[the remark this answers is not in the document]
astefanutti: I think we disable the workflow manually instead of introducing those changes. Alternatively I think that's acceptable we force merge the PR with the OLM workflow failing.

Dockerfile
dgrove-oss: Is there a RH registry that has the a usable go 1.21 toolchain that is publicly accessible?  I couldn't find one.  If there is one, I'd be happy to change the docker file.

astefanutti: That's fine for upstream, I think. This has to be overwritten anyway downstream.

README.md
```diff
@@ -90,7 +91,7 @@ The e2e tests can be executed locally by running the following commands:
 
 To properly run e2e tests on disconnected cluster user has to provide additional environment variables to properly configure testing environment:
 
-- `CODEFLARE_TEST_PYTORCH_IMAGE` - image tag for image used to run training job using MCAD
+- `CODEFLARE_TEST_PYTORCH_IMAGE` - image tag for image used to run training job using AppWrapper
```
astefanutti, line 94:
```suggestion
- `CODEFLARE_TEST_PYTORCH_IMAGE` - image tag for image used to run training job
```

config/rbac/kustomization.yaml
```diff
@@ -11,8 +11,6 @@ resources:
 - mcad_manager_role_binding.yaml
 - role.yaml
 - role_binding.yaml
-- instascale_role.yaml
-- instascale_role_binding.yaml
 - leader_election_role.yaml
 - leader_election_role_binding.yaml
 - mcad-controller-ray-clusterrolebinding.yaml
```
astefanutti, line 16: These two files can be removed right?

main.go
```diff
@@ -104,21 +97,26 @@ func main() {
...
+	namespace := namespaceOrDie()
 	cfg := &config.CodeFlareOperatorConfiguration{
 		ClientConnection: &config.ClientConnection{
-			QPS:   pointer.Float32(50),
-			Burst: pointer.Int32(100),
+			QPS:   ptr.To(float32(50)),
+			Burst: ptr.To(int32(100)),
+		},
+		AppWrapper: &config.AppWrapperConfiguration{
+			Enabled: ptr.To(true),
```
astefanutti, line 113: This should be disabled by default to start with.

main.go
```diff
@@ -191,8 +197,20 @@ func main() {
 		exitOnError(err, "Could not determine if RayCluster CR present on cluster.")
 	}
 
-	exitOnError(mgr.AddHealthzCheck(cfg.Health.LivenessEndpointName, healthz.Ping), "unable to set up health check")
-	exitOnError(mgr.AddReadyzCheck(cfg.Health.ReadinessEndpointName, healthz.Ping), "unable to set up ready check")
+	v1, err1 := HasAPIResourceForGVK(kubeClient.DiscoveryClient, kueue.GroupVersion.WithKind("Workload"))
+	v2, err2 := HasAPIResourceForGVK(kubeClient.DiscoveryClient, mcadv1beta2.GroupVersion.WithKind("AppWrapper"))
+	if v1 && v2 && *cfg.AppWrapper.Enabled {
+		// Ascynchronous because controllers need to wait for certificate to be ready for webhooks to work
```
astefanutti, line 203:
```suggestion
		// Asynchronous because controllers need to wait for certificate to be ready for webhooks to work
```

main.go
```diff
@@ -167,20 +182,11 @@ func main() {
...
-			Client: mgr.GetClient(),
-			Scheme: mgr.GetScheme(),
-			Config: cfg.InstaScale.InstaScaleConfiguration,
-		}
-		exitOnError(instaScaleController.SetupWithManager(context.Background(), mgr), "Error setting up InstaScale controller")
+	certsReady := make(chan struct{})
+	if os.Getenv("ENABLE_WEBHOOKS") == "false" {
+		close(certsReady)
+	} else {
+		exitOnError(awctrl.SetupCertManagement(mgr, &cfg.AppWrapper.Config.CertManagement, certsReady), "unable to set up cert rotation")
```
astefanutti, line 189: Can the setup of cert-controller be owned by the CodeFlare operator. It's going to be needed also for other controllers, so it's better `SetupCertManagement` to be in the operator itself.

.github/workflows/olm_tests.yaml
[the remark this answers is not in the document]
dgrove-oss: ok.  I'll revert the changes to this workflow and assume we'll ignore/override the failure to merge the PR.

main.go
dgrove-oss: Sure.  Makes sense.

main.go
dgrove-oss: Sure I can do that.  It might also allow us to merge this PR without waiting on SDK changes to enable the new AppWrappers (we'd need to be sure all the old MCAD code is disabled by default, but I think that has mostly been done already).

We should have a followup PR that adds e2e testing (another workflow?) in both this repo and the SDK that enables the AppWrapper controller though.

astefanutti: Yes exactly, that makes things easier w.r.t. the SDK.

dgrove-oss: Done.  I'll do a trial PR to the SDK that just makes the change to this branch of the operator but does the minimal other changes.

main.go
dgrove-oss: Done.

.github/workflows/olm_tests.yaml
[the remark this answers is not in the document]
dgrove-oss: kept the change to go version; reverted the rest of the file

main.go
```diff
@@ -167,20 +190,11 @@ func main() {
...
-			Client: mgr.GetClient(),
-			Scheme: mgr.GetScheme(),
-			Config: cfg.InstaScale.InstaScaleConfiguration,
-		}
-		exitOnError(instaScaleController.SetupWithManager(context.Background(), mgr), "Error setting up InstaScale controller")
+	certsReady := make(chan struct{})
+	if os.Getenv("ENABLE_WEBHOOKS") == "false" {
+		close(certsReady)
+	} else {
+		exitOnError(setupCertManagement(mgr, cfg.CertManagement, certsReady), "unable to set up cert rotation")
```
astefanutti, line 197: @KPostOffice FYI for the webhooks.

main.go
```diff
@@ -127,15 +123,22 @@ func main() {
 			},
 			LeaderElection: &configv1alpha1.LeaderElectionConfiguration{},
 		},
-		MCAD: &mcadconfig.MCADConfiguration{},
-		InstaScale: &config.InstaScaleConfiguration{
-			Enabled: pointer.Bool(false),
-			InstaScaleConfiguration: instascaleconfig.InstaScaleConfiguration{
-				MaxScaleoutAllowed: 5,
-			},
+		CertManagement: &config.CertManagementConfig{
```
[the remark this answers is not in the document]
dgrove-oss: With config.go and main.go being in separate packages, I don't think there is a natural way to do this if you want to be able to change it in the config map (which I was assuming you might?).  

If it's ok to be totally hardwired, then I can move the struct into `main.go` and make it private, but then I don't see how it gets changed.  Maybe I'm overlooking another option though.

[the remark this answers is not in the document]
astefanutti: Right I meant we don't need / want that in the ConfigMap. I'd rather be conservative and keep it hardwired to start with.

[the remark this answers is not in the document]
dgrove-oss: Got it.  Done.

.github/workflows/e2e_tests.yaml
```diff
@@ -82,7 +82,7 @@ jobs:
           echo "CODEFLARE_TEST_OUTPUT_DIR=${CODEFLARE_TEST_OUTPUT_DIR}" >> $GITHUB_ENV
 
           set -euo pipefail
-          go test -timeout 30m -v ./test/e2e -json 2>&1 | tee ${CODEFLARE_TEST_OUTPUT_DIR}/gotest.log | gotestfmt
+          go test -timeout 30m -v -skip ".*AppWrapper.*" ./test/e2e -json 2>&1 | tee ${CODEFLARE_TEST_OUTPUT_DIR}/gotest.log | gotestfmt
```
astefanutti, line 85: For the e2e tests, we can have a custom config that enables the AppWrapper controller.

dgrove-oss:
Replaced by #543.

PR #493, opened 2024-03-21
Refactor creation logic of ingress/routes into RayCluster Controller

ChristianZaccaria:
# Issue link

Issue: https://issues.redhat.com/browse/RHOAIENG-1056 

# What changes have been made

- Refactor the ingress/route creation and TLS support to the RayCluster Controller in the CFO.
- Added creation of rayclient route OpenShift, rayclient ingress and ingresses for KinD/vanilla clusters.
- RayCluster Controller to always run alongside the CFO.
- Adjusted e2e tests to disable OAuth in the CFO config, using this config bool value to indicate to the RayCluster Controller whether to create a secure route and OAuth resources on OpenShift or an ingress for KinD clusters (for e2e tests).
- Created a support package and several functions including:
  - `createRayClientRoute`
  - `createRayClientIngress`
  - `createIngressApplyConfiguration`
  - `isOnKindCluster`
  - `getClusterType`
  - `annotationBoolVal`
  - `getIngressDomain`

# Verification steps

This PR should be tested alongside this PR in the SDK which removes the creation of ingress/routes creation logic: https://github.com/project-codeflare/codeflare-sdk/pull/495 
To test these changes:
**In an OpenShift cluster:**

1. Checkout this PR and run `make run NAMESPACE=default` and `make install`.
2. Deploy and install KubeRay `v1.1.0`
3. Checkout this PR and install the SDK: https://github.com/project-codeflare/codeflare-sdk/pull/495
4. Deploy Kueue and create a `LocalQueue` and `ClusterQueue`.
5. Run through [this notebook](https://github.com/project-codeflare/codeflare-sdk/blob/403cca63c759eb43cc0279cae04f91b83fc1fb55/demo-notebooks/additional-demos/local_interactive.ipynb)
6. Switch `mcad` to false.
7. Add `write_to_file` to be true to have access to the RayCluster yaml and add the Kueue label `kueue.x-k8s.io/queue-name: <name of localqueue>`
8. Add an `ingress_domain` to the ClusterConfiguration i.e., `apps.rosa.clustername.k1pm.p3.openshiftapps.com`
9. Once cluster.up() is ran, the oauth-proxy sidecar is created by default on the SDK side if on an OpenShift cluster and the RayCluster Controller creates all the necessary OAuth resources + necessary rayclient route if the RayCluster is not in a `suspended` state provided by Kueue. If not enough resources, The RayCluster controller will hold the creation of the routes/ingresses.
10. After going through the notebook, the route should be accessible through authentication and running `ray.get(ref)` should result in = 1789.4644.....

**In a KinD cluster:**
1. For the local_interactive notebook, same steps as above except you will find that no OAuth resources are created including the oauth-proxy sidecar container. The ingress and RayClient ingress will be created by the RayCluster Controller.
- Add in your local /etc/hosts `127.0.0.1 kind` [as per documentation](https://github.com/project-codeflare/codeflare-sdk/blob/main/docs/e2e.md#:~:text=Pre%2Drequisite%20for%20KinD%20clusters%3A%20please%20add%20in%20your%20local%20/etc/hosts%20file%20127.0.0.1%20kind)
- For the ingress_domain, use `kind` unless you are running an un-named KinD cluster of which hostname will resolve to `kind-control-plane` by default, in this scenario, `ingress_domain` is not required.
- After going through the notebook, the route should be accessible through authentication and running `ray.get(ref)` should result in = 1789.4644.....

2. For a [basic notebook](https://github.com/project-codeflare/codeflare-sdk/blob/main/demo-notebooks/guided-demos/2_basic_jobs.ipynb), same steps involved as above, except you will find that the RayClient ingress is not created this time, only the ingress to the service is created.

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/.github/workflows/e2e_tests.yaml
+++ b/.github/workflows/e2e_tests.yaml
@@ -60,6 +60,7 @@ jobs:
         run: |
           echo Deploying CodeFlare operator
           IMG="${REGISTRY_ADDRESS}"/codeflare-operator
+          sed -i 's/RayDashboardOAuthEnabled: pointer.Bool(true)/RayDashboardOAuthEnabled: pointer.Bool(false)/' main.go
           make image-push -e IMG="${IMG}"
           make deploy -e IMG="${IMG}" -e ENV="e2e"
           kubectl wait --timeout=120s --for=condition=Available=true deployment -n openshift-operators codeflare-operator-manager
--- a/main.go
+++ b/main.go
@@ -188,7 +188,7 @@ func main() {
 	}
 
 	v, err := HasAPIResourceForGVK(kubeClient.DiscoveryClient, rayv1.GroupVersion.WithKind("RayCluster"))
-	if v && *cfg.KubeRay.RayDashboardOAuthEnabled {
+	if v {
 		rayClusterController := controllers.RayClusterReconciler{Client: mgr.GetClient(), Scheme: mgr.GetScheme()}
 		exitOnError(rayClusterController.SetupWithManager(mgr), "Error setting up RayCluster controller")
 	} else if err != nil {
--- a/main.go
+++ b/main.go
@@ -189,7 +189,7 @@ func main() {
 
 	v, err := HasAPIResourceForGVK(kubeClient.DiscoveryClient, rayv1.GroupVersion.WithKind("RayCluster"))
 	if v {
-		rayClusterController := controllers.RayClusterReconciler{Client: mgr.GetClient(), Scheme: mgr.GetScheme()}
+		rayClusterController := controllers.RayClusterReconciler{Client: mgr.GetClient(), Scheme: mgr.GetScheme(), Config: cfg}
 		exitOnError(rayClusterController.SetupWithManager(mgr), "Error setting up RayCluster controller")
 	} else if err != nil {
 		exitOnError(err, "Could not determine if RayCluster CR present on cluster.")
--- a/pkg/controllers/raycluster_controller.go
+++ b/pkg/controllers/raycluster_controller.go
@@ -52,6 +52,7 @@ type RayClusterReconciler struct {
 	routeClient *routev1client.RouteV1Client
 	Scheme      *runtime.Scheme
 	CookieSalt  string
+	Config      *config.CodeFlareOperatorConfiguration
 }
 
 const (
@@ -60,14 +61,13 @@ const (
 	oAuthFinalizer         = "ray.openshift.ai/oauth-finalizer"
 	oAuthServicePort       = 443
 	oAuthServicePortName   = "oauth-proxy"
-	regularServicePortName = "dashboard"
+	ingressServicePortName = "dashboard"
 	logRequeueing          = "requeueing"
 )
 
 var (
-	deletePolicy   = metav1.DeletePropagationForeground
-	deleteOptions  = client.DeleteOptions{PropagationPolicy: &deletePolicy}
-	configInstance *config.CodeFlareOperatorConfiguration
+	deletePolicy  = metav1.DeletePropagationForeground
+	deleteOptions = client.DeleteOptions{PropagationPolicy: &deletePolicy}
 )
 
 // +kubebuilder:rbac:groups=ray.io,resources=rayclusters,verbs=get;list;watch;create;update;patch;delete
@@ -101,9 +101,9 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	isLocalInteractive := annotationBoolVal(logger, &cluster, "sdk.codeflare.dev/local_interactive")
-	isOpenShift, ingressHost := getClusterType(logger, r.kubeClient, &cluster)
-	ingressDomain := cluster.ObjectMeta.Annotations["sdk.codeflare.dev/ingress_domain"]
+	isLocalInteractive := annotationBoolVal(ctx, &cluster, "sdk.codeflare.dev/local_interactive", false)
+	ingressDomain := "" // FIX - CFO will retrieve it.
+	isOpenShift, ingressHost := getClusterType(ctx, r.kubeClient, &cluster, ingressDomain)
 
 	if cluster.ObjectMeta.DeletionTimestamp.IsZero() {
 		if !controllerutil.ContainsFinalizer(&cluster, oAuthFinalizer) {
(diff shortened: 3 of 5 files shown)


pkg/controllers/raycluster_controller.go
```diff
@@ -130,29 +138,56 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 
-	_, err := r.routeClient.Routes(cluster.Namespace).Apply(ctx, desiredClusterRoute(&cluster), metav1.ApplyOptions{FieldManager: controllerName, Force: true})
-	if err != nil {
-		logger.Error(err, "Failed to update OAuth Route")
-	}
+	if cluster.Status.State != "suspended" && r.isRayDashboardOAuthEnabled() && isOpenShift {
+		logger.Info("Creating OAuth Objects")
+		_, err := r.routeClient.Routes(cluster.Namespace).Apply(ctx, desiredClusterRoute(&cluster), metav1.ApplyOptions{FieldManager: controllerName, Force: true})
+		if err != nil {
+			logger.Error(err, "Failed to update OAuth Route")
```
anishasthana, line 145: I _think_ we should be returning the error with all of these failures. Similar to what Edson mentioned/proposed in https://github.com/project-codeflare/codeflare-operator/pull/498/files.

`return ctrl.Result{RequeueAfter: requeueTime}, err`

pkg/controllers/raycluster_controller.go
```diff
@@ -97,6 +101,10 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	isLocalInteractive := annotationBoolVal(logger, &cluster, "sdk.codeflare.dev/local_interactive")
+	isOpenShift, ingressHost := getClusterType(logger, r.kubeClient, &cluster)
+	ingressDomain := getIngressDomain(&cluster)
```
KPostOffice, line 106: Since you only care about the presence of ingress

This can just be simplified to:
```
_, ingressDomainExists := cluster.ObjectMeta.Annotations["sdk.codeflare.dev/ingress_domain"]
```
and edit the conditionals below

pkg/controllers/support.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// getDiscoveryClient returns a discovery client for the current reconciler
+func getDiscoveryClient(config *rest.Config) (*discovery.DiscoveryClient, error) {
+	return discovery.NewDiscoveryClientForConfig(config)
+}
+
+// Check where we are running. We are trying to distinguish here whether
+// this is vanilla kubernetes cluster or Openshift
+func getClusterType(logger logr.Logger, clientset *kubernetes.Clientset, cluster *rayv1.RayCluster) (bool, string) {
```
KPostOffice, line 132: Small nit. Can you pass ctx to these functions instead and get logger via ctrl.loggerfrom

pkg/controllers/support.go
```diff
@@ -0,0 +1,194 @@
...
+				}
+			}
+		} else {
+			logger.Info("Cannot retrieve a DiscoveryClient, assuming we're on Vanilla Kubernetes")
+			return false, fmt.Sprintf("ray-dashboard-%s-%s.%s", cluster.Name, cluster.Namespace, ingress_domain)
+		}
+	} else {
+		logger.Info("Cannot retrieve config, assuming we're on Vanilla Kubernetes")
+		return false, fmt.Sprintf("ray-dashboard-%s-%s.%s", cluster.Name, cluster.Namespace, ingress_domain)
+	}
```
KPostOffice, line 166: Can you change this to:
```
if err != nil || dclient == nil {
    logger.Info("Cannot retrieve config, assuming we're on Vanilla Kubernetes")
	return false, fmt.Sprintf("ray-dashboard-%s-%s.%s", cluster.Name, cluster.Namespace, ingress_domain)
}
```
and then leave the majority of the body outside of conditionals

#NeverNester https://www.youtube.com/watch?v=CFRhGnuXG-4

KPostOffice: Oops wrong conditional, but same with this one:
```
if err != nil || config == nil {
    ...
    return ...
}
```

pkg/controllers/support.go
```diff
@@ -0,0 +1,194 @@
...
+	ingress_domain := cluster.ObjectMeta.Annotations["sdk.codeflare.dev/ingress_domain"]
+	config, err := ctrl.GetConfig()
+	if err == nil && config != nil {
+		dclient, err := getDiscoveryClient(config)
+		if err == nil && dclient != nil {
+			apiGroupList, err := dclient.ServerGroups()
+			if err != nil {
+				logger.Info("Error while querying ServerGroups, assuming we're on Vanilla Kubernetes")
+				return false, ""
+			} else {
```
KPostOffice, line 143: And here too but just unindent since it is already unreachable due to `return`

#NeverNest!

pkg/controllers/support.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+func (r *RayClusterReconciler) isRayDashboardOAuthEnabled() bool {
+	if r.Config != nil && r.Config.KubeRay != nil && r.Config.KubeRay.RayDashboardOAuthEnabled != nil {
+		return *r.Config.KubeRay.RayDashboardOAuthEnabled
+	}
+	return true
+}
+
+func annotationBoolVal(logger logr.Logger, cluster *rayv1.RayCluster, annotation string) bool {
```
KPostOffice, line 176: Can you add another arg for default value in case default should be `true` in some future code?

pkg/controllers/support.go
```diff
@@ -0,0 +1,194 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	routeapply "github.com/openshift/client-go/route/applyconfigurations/route/v1"
+)
+
+func serviceNameFromCluster(cluster *rayv1.RayCluster) string {
+	return cluster.Name + "-head-svc"
+}
+
+func createRayClientRoute(cluster *rayv1.RayCluster) *routeapply.RouteApplyConfiguration {
```
KPostOffice, line 30: I think `create` is a misnomer here. Can you use the same verbage use for the other resources (`desired`)?

pkg/controllers/support.go
ChristianZaccaria: Wow Kevin that video was so enlightening! Thanks for saving me early, I will 100% follow the path of a Never Nester. Hopefully this next commit shows that :) - Thanks a lot for the video and advise.

pkg/controllers/support.go
ChristianZaccaria: That makes complete sense, thanks!!

pkg/controllers/raycluster_controller.go
ChristianZaccaria: That makes sense to me, this way we can fail-fast and re-queue. Change has been made. - Thanks Anish!

ChristianZaccaria (commented):
Thanks for the review. I added a commit per comment - Will squash them afterwards.

pkg/controllers/raycluster_controller.go
ChristianZaccaria: I see why that can be implied, that was wrong on my part. I do require the actual value of the ingress domain annotation too as it's used for the Host field for creating ingresses and RayClient ingress/route.

I was originally declaring a new variable for ingress domain on each creation of ingresses, that's why it looks like I only needed the presence of it. Great catch thanks. Changes made.

pkg/controllers/support.go
ChristianZaccaria: Wow Kevin that video was so enlightening! Thanks for saving me early, I will 100% follow the path of a Never Nester. Hopefully this next commit shows that :) - Thanks a lot for the video and advise.

pkg/controllers/support.go
ChristianZaccaria: That makes complete sense, thanks!!

pkg/controllers/support.go
ChristianZaccaria: Good call, thanks!

.github/workflows/e2e_tests.yaml
```diff
@@ -60,6 +60,7 @@ jobs:
         run: |
           echo Deploying CodeFlare operator
           IMG="${REGISTRY_ADDRESS}"/codeflare-operator
+          sed -i 's/RayDashboardOAuthEnabled: pointer.Bool(true)/RayDashboardOAuthEnabled: pointer.Bool(false)/' main.go
```
astefanutti, line 63: That should be done in the e2e ConfigMap instead.

main.go
```diff
@@ -184,8 +184,8 @@ func main() {
 	}
 
 	v, err := HasAPIResourceForGVK(kubeClient.DiscoveryClient, rayv1.GroupVersion.WithKind("RayCluster"))
-	if v && *cfg.KubeRay.RayDashboardOAuthEnabled {
-		rayClusterController := controllers.RayClusterReconciler{Client: mgr.GetClient(), Scheme: mgr.GetScheme()}
+	if v {
```
astefanutti, line 187: nit: Let's rename that `v` to `ok` :)

pkg/controllers/raycluster_controller.go
```diff
@@ -50,15 +52,17 @@ type RayClusterReconciler struct {
 	routeClient *routev1client.RouteV1Client
 	Scheme      *runtime.Scheme
 	CookieSalt  string
+	Config      *config.CodeFlareOperatorConfiguration
```
astefanutti, line 55: Maybe only the RayCluster controller configuration could be enough here?

pkg/controllers/raycluster_controller.go
```diff
@@ -97,6 +101,10 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	isLocalInteractive := annotationBoolVal(ctx, &cluster, "sdk.codeflare.dev/local_interactive", false)
+	ingressDomain := cluster.ObjectMeta.Annotations["sdk.codeflare.dev/ingress_domain"]
+	isOpenShift, ingressHost := getClusterType(ctx, r.kubeClient, &cluster, ingressDomain)
```
astefanutti, line 106: Checking the cluster type should only be done once when the operator starts. Calling the Discovery API for each reconciliation is also not really acceptable.

pkg/controllers/raycluster_controller.go
```diff
@@ -130,29 +138,63 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, nil
 	}
 
-	_, err := r.routeClient.Routes(cluster.Namespace).Apply(ctx, desiredClusterRoute(&cluster), metav1.ApplyOptions{FieldManager: controllerName, Force: true})
-	if err != nil {
-		logger.Error(err, "Failed to update OAuth Route")
-	}
+	if cluster.Status.State != "suspended" && r.isRayDashboardOAuthEnabled() && isOpenShift {
```
astefanutti, line 141: It's possible the cluster transitions from running to suspended state. Should the resources be removed in that case?

pkg/controllers/raycluster_controller.go
```diff
@@ -97,6 +101,10 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	isLocalInteractive := annotationBoolVal(ctx, &cluster, "sdk.codeflare.dev/local_interactive", false)
+	ingressDomain := cluster.ObjectMeta.Annotations["sdk.codeflare.dev/ingress_domain"]
```
astefanutti, line 105: What does justify we let clients / users decide for these? Shouldn't that be the responsibility of the platform admin to configure?

ChristianZaccaria: Those are the annotations that are created when local_interactive is true or the user is using an ingress domain.
The local_interactive one is used for the RCC know to create the Ray Client.
And the ingress domain one is needed for Kubernetes clusters right?

pkg/controllers/support.go
```diff
@@ -0,0 +1,181 @@
...
+		logger.Info("Error while querying ServerGroups, assuming we're on Vanilla Kubernetes")
+		return false, ""
+	}
+	for i := 0; i < len(apiGroupList.Groups); i++ {
+		if strings.HasSuffix(apiGroupList.Groups[i].Name, ".openshift.io") {
+			logger.Info("We detected being on OpenShift!")
+			return true, ""
+		}
+	}
+	onKind, _ := isOnKindCluster(clientset)
```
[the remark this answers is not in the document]
ChristianZaccaria: You are right, we don't explicitly need to check for KinD, anything that is not OpenShift could suit. The only reason that we check for KinD is for our own testing purposes. If it's true, then the ingress Host will use "kind". We could make it more generic by supplying the ingress_domain to the e2e tests, then there is no need for checking for KinD explicitly. Should I change or leave as is for now... WDYT?

Bobbins228 (commented):
/lgtm
Just tested it out and it creates oauth objects successfully and does not when the RC is suspended.
Also tested suspended -> admitted working as expected

sutaakar:
I miss here a test coverage being added into https://github.com/project-codeflare/codeflare-operator/blob/faff28ab80941db913f509ffdc22eb9cffbed7c8/pkg/controllers/raycluster_controller_test.go

PR #495, opened 2024-03-28
Fix Unknown build info in CodeFlare operator log

oksanabaza:
# Issue link

https://issues.redhat.com/browse/RHOAIENG-118
# What changes have been made

Fix Unknown build info in CodeFlare operator log
# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,11 +91,12 @@ endif
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
-BUILD_DATE := $(shell date +%Y-%m-%d\ %H:%M)
+export BUILD_DATE := $(shell date +%Y-%m-%d\ %H:%M)
+
 BUILD_TAG_SHA := $(shell git rev-list --abbrev-commit --tags --max-count=1)
 BUILD_TAG_NAME := $(shell git describe --abbrev=0 --tags ${BUILD_TAG_SHA} 2>/dev/null || true)
 BUILD_SHA := $(shell git rev-parse --short HEAD)
-BUILD_VERSION := $(BUILD_TAG_NAME:v%=%)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -174,7 +174,14 @@ build: fmt vet ## Build manager binary.
...
 run: manifests fmt vet ## Run a controller from your host.
-	go run ./main.go
+	go run \
+		-ldflags " \
+			-X 'main.OperatorVersion=$(BUILD_VERSION)' \
+			-X 'main.McadVersion=$(MCAD_VERSION)'  \
+			-X 'main.InstaScaleVersion=$(INSTASCALE_VERSION)' \
+			-X 'main.BuildDate=$(BUILD_DATE)' \
+		" \
+		./main.go
```
KPostOffice, line 184: So this will only set these values when `make run` is used to run the code locally. I believe the root of this issue is actually the Dockerfile. It directly calls `go build` rather than using the make directive specified in this file. 

If you change the dockerfile to use `make build` https://github.com/project-codeflare/codeflare-operator/pull/495/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R164-R173 I think it will solve the issue.

config/manager/params.env
```diff
@@ -1 +1 @@
-codeflare-operator-controller-image=quay.io/opendatahub/codeflare-operator:v1.2.0
+codeflare-operator-controller-image=quay.io/rh-ee-obazylie/codeflare-operator:test1
```
KPostOffice, line 1: Left over testing val

KPostOffice (commented):
Thanks @oksanabaza !

PR #498, opened 2024-04-03
add additional function which creates network policy

KPostOffice:
# Issue link

# What changes have been made

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


pkg/controllers/raycluster_controller.go
```diff
@@ -154,6 +156,11 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		logger.Error(err, "Failed to update OAuth ClusterRoleBinding")
 	}
 
+	_, err = r.kubeClient.NetworkingV1().NetworkPolicies(cluster.Namespace).Apply(ctx, desiredNetworkPolicy(&cluster), metav1.ApplyOptions{FieldManager: controllerName, Force: true})
+	if err != nil {
+		logger.Error(err, "Failed to update NetworkPolicy")
+	}
```
etirelli, line 162: shouldn't the error be returned? I see the same is done for the previous objects (ClusterRoleBindings, ServiceAccounts, etc)

etirelli (approved):
I am approving it, but please take a look at my comment. It seems strange to log the error but not return a failure in the reconciliation.

pkg/controllers/raycluster_controller.go
```diff
@@ -267,6 +274,36 @@ func desiredOAuthSecret(cluster *rayv1.RayCluster, r *RayClusterReconciler) *cor
...
+	return networkingapply.NetworkPolicy(cluster.Name, cluster.Namespace).
+		WithLabels(map[string]string{"ray.io/cluster-name": cluster.Name}).
+		WithSpec(networkingapply.NetworkPolicySpec().
+			WithPodSelector(metav1apply.LabelSelector().WithMatchLabels(map[string]string{"ray.io/cluster": cluster.Name, "ray.io/node-type": "head"})).
+			WithIngress(
+				networkingapply.NetworkPolicyIngressRule().
+					WithPorts(
+						networkingapply.NetworkPolicyPort().WithProtocol(corev1.ProtocolTCP).WithPort(intstr.FromInt(6379)),
+						networkingapply.NetworkPolicyPort().WithProtocol(corev1.ProtocolTCP).WithPort(intstr.FromInt(10001)),
+						networkingapply.NetworkPolicyPort().WithProtocol(corev1.ProtocolTCP).WithPort(intstr.FromInt(8080)),
```
astefanutti, line 288: The metrics endpoint should be accessible from the OpenShift user workload monitoring namespace. That should be added in a separate rule.

pkg/controllers/raycluster_controller.go
```diff
@@ -267,6 +274,36 @@ func desiredOAuthSecret(cluster *rayv1.RayCluster, r *RayClusterReconciler) *cor
 	// Create a Kubernetes secret to store the cookie secret
 }
 
+func desiredNetworkPolicy(cluster *rayv1.RayCluster) *networkingapply.NetworkPolicyApplyConfiguration {
+
+	return networkingapply.NetworkPolicy(cluster.Name, cluster.Namespace).
```
zdtsw, line 279: is it easier to use `"k8s.io/api/networking/v1"` instead of `"k8s.io/client-go/applyconfigurations/networking/v1"`
```
desiredNetworkPolicy := &netv1.NetworkPolicy{
			TypeMeta: metav1.TypeMeta{
				Kind:       "NetworkPolicy",
				APIVersion: "v1",
			},
			ObjectMeta: metav1.ObjectMeta{
				Name:      name,
				Namespace: name,
				Labels:    map[string]string{"ray.io/cluster": cluster.Name, "ray.io/node-type": "head"},
			},
			Spec: netv1.NetworkPolicySpec{
				Ingress: []netv1.NetworkPolicyIngressRule{
					{
						Ports: []netv1.NetworkPolicyPort{
							{
								Port: &intstr.IntOrString{Type: intstr.Int, IntVal: 6379},
							},
							{
								Port: &intstr.IntOrString{Type: intstr.Int, IntVal: 10001},
							},
						},
						From: []netv1.NetworkPolicyPeer{
							{
								NamespaceSelector: &metav1.LabelSelector{
									MatchLabels: map[string]string{
										"kubernetes.io/metadata.name": "value of application namespace",
									},
								},
								PodSelector: &metav1.LabelSelector{
									MatchLabels: map[string]string{
										"app.kubernetes.io/component": "kuberay-operator",
									},
								},
							},
						},
					},
					{
						Ports: []netv1.NetworkPolicyPort{
							{
								Port: &intstr.IntOrString{Type: intstr.Int, IntVal: 8443},
							},
						},
					},
				},
				PolicyTypes: []netv1.PolicyType{
					netv1.PolicyTypeIngress,
				},
			},
		}
```

astefanutti: That's needed for SSA, that brings other advantages, like one-shot server-side patching, conflict resolutions with concurrent controllers like KubeRay, etc.

zdtsw: thanks for the tips!
i think we will need to move towards this way in operator's  code as well.

astefanutti: Right, from my experience, SSA makes things much simpler from a code standpoint and more robust from a runtime one.

pkg/controllers/raycluster_controller.go
```diff
@@ -267,6 +274,41 @@ func desiredOAuthSecret(cluster *rayv1.RayCluster, r *RayClusterReconciler) *cor
...
+						networkingapply.NetworkPolicyPort().WithProtocol(corev1.ProtocolTCP).WithPort(intstr.FromInt(10001)),
+						networkingapply.NetworkPolicyPort().WithProtocol(corev1.ProtocolTCP).WithPort(intstr.FromInt(8080)),
+						networkingapply.NetworkPolicyPort().WithProtocol(corev1.ProtocolTCP).WithPort(intstr.FromInt(8265)),
+					).WithFrom(
+					networkingapply.NetworkPolicyPeer().WithPodSelector(metav1apply.LabelSelector()),
+				),
+				networkingapply.NetworkPolicyIngressRule().WithFrom(
+					networkingapply.NetworkPolicyPeer().WithPodSelector(metav1apply.LabelSelector().
+						WithMatchLabels(map[string]string{"app.kubernetes.io/component": "kuberay-operator"})).
+						WithNamespaceSelector(metav1apply.LabelSelector().WithMatchLabels(map[string]string{"opendatahub.io/generated-namespace": "true"})),
```
astefanutti, line 296: I don't think relying on the `opendatahub.io/generated-namespace` label is correct. The ingress traffic should be allowed from the ODH / RHOAI application namespace. So it should either be sourced from the DSCInitilization resource, or defaulted to the well-known namespaces

zdtsw: should we hardcode it `opendatahub` for ODH and `redhat-ods-applications` for RHOAI to simplify the logic for now? 
and think about how to get it from DSCI later? 
i mentioned it in another thread, operator does make this namespace configurable, but unlikely in ODH user will change the default value

pkg/controllers/raycluster_controller.go
```diff
@@ -197,6 +201,25 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 	}
 
+	// Locate the KubeRay operator deployment:
+	// - First try to get the ODH / RHOAI application namespace from the DSCInitialization
+	// - Or fallback to the well-known defaults
+	var kubeRayNamespaces []string
+	dsci := &dsciv1.DSCInitialization{}
+	err := r.Client.Get(ctx, client.ObjectKey{Name: "default-dsci"}, dsci)
```
sutaakar, line 209: Can you load DSCI without hardcoding name? This will make code more resilient for possible naming changes.

zdtsw: default-dsci is the one created by ODH/RHOAI operator.
99% case i would expect it is using this name, but ofc if user tried to create CR by themselves, that will do.
for the time being, we could hardcode with this name. 
a follow up PR can add more logic "check CR, get name, set here"

main.go
```diff
@@ -70,6 +71,8 @@ func init() {
 	utilruntime.Must(rayv1.AddToScheme(scheme))
 	// OpenShift Route
 	utilruntime.Must(routev1.Install(scheme))
+	// ODH
+	utilruntime.Must(odhv1.AddToScheme(scheme))
```
sutaakar, line 75: Got error:
```
2024-04-15T13:11:03+02:00	ERROR	Reconciler error	{"controller": "codeflare-raycluster-controller", "controllerGroup": "ray.io", "controllerKind": "RayCluster", "RayCluster": {"name":"jobtest","namespace":"ksuta-test"}, "namespace": "ksuta-test", "name": "jobtest", "reconcileID": "3f44de47-9706-40d1-a470-62a4db9869dc", "error": "no kind is registered for the type v1.DSCInitialization in scheme \"pkg/runtime/scheme.go:100\""}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/home/ksuta/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.3/pkg/internal/controller/controller.go:324
```

Was able to fix it by using scheme from `dsciv1 "github.com/opendatahub-io/opendatahub-operator/v2/apis/dscinitialization/v1"`

pkg/controllers/raycluster_controller.go
franciscojavierarceo: Just reviewing this and maybe the comment got moved around but was this addressed?

pkg/controllers/raycluster_controller.go
```diff
@@ -337,6 +361,56 @@ func desiredOAuthSecret(cluster *rayv1.RayCluster, r *RayClusterReconciler) *cor
...
+				networkingapply.NetworkPolicyIngressRule().
+					WithPorts(
+						networkingapply.NetworkPolicyPort().WithProtocol(corev1.ProtocolTCP).WithPort(intstr.FromInt(8080)),
+					).
+					WithFrom(
+						networkingapply.NetworkPolicyPeer().WithNamespaceSelector(metav1apply.LabelSelector().
+							WithMatchExpressions(metav1apply.LabelSelectorRequirement().
+								WithKey(corev1.LabelMetadataName).
+								WithOperator(metav1.LabelSelectorOpIn).
+								WithValues("openshift-user-workload-monitoring"))),
```
astefanutti, line 401: I think that should be `openshift-monitoring` instead.

zdtsw: does SRE monitoring need access redhat-ods-applications (your ray controller) ? or this is for UWM?

pkg/controllers/raycluster_controller.go
```diff
@@ -337,6 +361,56 @@ func desiredOAuthSecret(cluster *rayv1.RayCluster, r *RayClusterReconciler) *cor
...
+		WithSpec(networkingapply.NetworkPolicySpec().
+			WithPodSelector(metav1apply.LabelSelector().WithMatchLabels(map[string]string{"ray.io/cluster": cluster.Name, "ray.io/node-type": "head"})).
+			WithIngress(
+				networkingapply.NetworkPolicyIngressRule().
+					WithPorts(
+						networkingapply.NetworkPolicyPort().WithProtocol(corev1.ProtocolTCP).WithPort(intstr.FromInt(6379)),
+						networkingapply.NetworkPolicyPort().WithProtocol(corev1.ProtocolTCP).WithPort(intstr.FromInt(10001)),
+						networkingapply.NetworkPolicyPort().WithProtocol(corev1.ProtocolTCP).WithPort(intstr.FromInt(8080)),
+						networkingapply.NetworkPolicyPort().WithProtocol(corev1.ProtocolTCP).WithPort(intstr.FromInt(8265)),
+					).WithFrom(
```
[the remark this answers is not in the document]
KPostOffice: If you have an empty pod selector it selects all pods from the same namespace. It is the same as disallowing access from other namespaces.

https://github.com/ahmetb/kubernetes-network-policy-recipes/blob/master/04-deny-traffic-from-other-namespaces.md

pkg/controllers/raycluster_controller.go
KPostOffice: I was planning to open another PR that fixes all the instances of this issue once this is merged

PR #507, opened 2024-04-11
add: raycluster oauth mutating webhook

VanillaSpoon:
# Issue link
https://issues.redhat.com/browse/RHOAIENG-1991

# What changes have been made
Add Mutating Webhook to add OAuth SideCar to RayCluster

# Verification steps
Setup environment.
Run the Codeflare-Operator from main
Build and deploy raycluster_webhook.
- From within src:
  - podman build <quay.io/user/repo:version> .
  - podman push <quay.io/user/repo:version
- From within config
  -kubectl apply -f deployment.yaml

Ensure the webhook is running in default namespace (will be changing this)

From the sdk, create a raycluster
- Ensure the raycluster is deployed correctly
- Inspect the raycluster yaml to enusre the oauth configuraitons have been applied
- Inspect the raycluster head node and ensure the oauth is running alongside it.

## Checks
- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [ ] Testing is not required for this change

PR #508, opened 2024-04-11
[WIP] Add Mtls patch to Mutating Webhook 

Bobbins228:
# Issue link

# What changes have been made

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


raycluster_webhook/src/patches.go
```diff
@@ -0,0 +1,352 @@
...
+		{
+			Name:      "server-cert",
+			MountPath: "/home/ray/workspace/tls",
+			ReadOnly:  false,
+		},
+	}
+	svcDomain := rayCluster.Name + "-head-svc." + rayCluster.Namespace + ".svc"
+	secretName := `ca-secret-` + rayCluster.Name
+	fmt.Println("Start")
+	domain, err := getDomainName()
```
franciscojavierarceo, line 144: We know the error starts here.

raycluster_webhook/src/patches.go
```diff
@@ -0,0 +1,352 @@
...
+		fmt.Errorf("Could not convert annotation value to bool", "annotation", annotation, "value", val)
+		return defaultValue
+	}
+	return boolVal
+}
+
+func getDomainName() (string, error) {
+	consoleRoute := &routev1.Route{}
+	fmt.Println("1")
+	fmt.Printf(client.ObjectKey{Name: NameConsoleLink, Namespace: NamespaceConsoleLink}.String())
```
franciscojavierarceo, line 341: and that this technically works, which I suppose isn't terribly shocking.

raycluster_webhook/src/patches.go
```diff
@@ -0,0 +1,352 @@
...
+	fmt.Println("2")
+	fmt.Printf(consoleRoute.String())
+
+	if err := k8Client.Get(context.TODO(), types.NamespacedName{Name: NameConsoleLink, Namespace: NamespaceConsoleLink}, consoleRoute); err != nil {
```
franciscojavierarceo, line 345: But this fails.

So, something is misbehaving into this `.Get()`.

1. It's probably not the `context.TODO()` 
2. I think we fixed the `types.NamespacedName` Struct but we should definitely try to confirm that.
3. The `consoleRoute` seemed reasonable from what I recall as that was the object that was passed to be updated.

We should definitely confirm these 3 things though and I'd dig into them slightly differently.

sutaakar: What is the error?
Do you run the container locally, connecting to the remote OpenShift cluster, or directly as a Deployment on the cluster?
If the latter, I see that you doesn't provide RBAC for accessing Routes - https://github.com/project-codeflare/codeflare-operator/pull/508/files#diff-d44acd75643a4ccbb2ddc671e641c9f520d495dbc01fe1dbcf305bd4c13915d9R53-R61

Bobbins228: @sutaakar This was the error we were receiving. I will update RBAC:
`no kind is registered for the type v1.Route in scheme "go/pkg/mod/k8s.io/apimachinery@v0.29.3/pkg/runtime/scheme.go:100"`

zdtsw:
not sure which one need review either #507 or #508 ? or none :scream: 
better remove [WIP] / "draft" and have the other one closed.

PR #512, opened 2024-04-12
Exit CFO initialisation if RayCluster CRD is not available

ChristianZaccaria:
# Issue link

Jira: https://issues.redhat.com/browse/RHOAIENG-5331 

# What changes have been made

- Refactored to exit early if the RayCluster CRD is not present in the cluster.

Note: This could be considered a temporary change until the RC Controller is moved to KubeRay.

# Verification steps

1. Deploy the CFO without KubeRay and Ray CRDs:
```
podman build -t quay.io/<quayusername>/codeflare-operator:<tagname> .
podman push quay.io/<quayusername>/codeflare-operator:<tagname>
make deploy IMG=quay.io/<quayusername>/codeflare-operator:<tagname>
```
2. The CFO pod will fail to initialise and will restart the pod until the required CRDs are available in the cluster. Deploying KubeRay installs the required CRDs and the CFO pod should start.

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/main.go
+++ b/main.go
@@ -128,8 +128,16 @@ func main() {
 	kubeClient, err := kubernetes.NewForConfig(kubeConfig)
 	exitOnError(err, "unable to create Kubernetes client")
 
-	namespace := namespaceOrDie()
+	// Checking for RayCluster CRD availability to ensure the RayCluster Controller can initialise.
+	ok, err := hasAPIResourceForGVK(kubeClient.DiscoveryClient, rayv1.GroupVersion.WithKind("RayCluster"))
+	if err != nil {
+		exitOnError(err, "Could not determine if RayCluster CR is present on cluster.")
+	}
+	if !ok {
+		exitOnError(fmt.Errorf("RayCluster CRD is not available on the cluster"), "Required CRD not available.")
+	}
 
+	namespace := namespaceOrDie()
 	exitOnError(loadIntoOrCreate(ctx, kubeClient, namespace, configMapName, cfg), "unable to initialise configuration")
 
 	kubeConfig.Burst = int(ptr.Deref(cfg.ClientConnection.Burst, int32(rest.DefaultBurst)))
@@ -171,18 +179,13 @@ func setupControllers(mgr ctrl.Manager, dc discovery.DiscoveryInterface, cfg *co
 
 	exitOnError(controllers.SetupRayClusterWebhookWithManager(mgr, cfg.KubeRay), "error setting up RayCluster webhook")
 
-	ok, err := hasAPIResourceForGVK(dc, rayv1.GroupVersion.WithKind("RayCluster"))
-	if ok {
-		rayClusterController := controllers.RayClusterReconciler{
-			Client:      mgr.GetClient(),
-			Scheme:      mgr.GetScheme(),
-			Config:      cfg.KubeRay,
-			IsOpenShift: isOpenShift,
-		}
-		exitOnError(rayClusterController.SetupWithManager(mgr), "Error setting up RayCluster controller")
-	} else if err != nil {
-		exitOnError(err, "Could not determine if RayCluster CR present on cluster.")
+	rayClusterController := controllers.RayClusterReconciler{
+		Client:      mgr.GetClient(),
+		Scheme:      mgr.GetScheme(),
+		Config:      cfg.KubeRay,
+		IsOpenShift: isOpenShift,
 	}
+	exitOnError(rayClusterController.SetupWithManager(mgr), "Error setting up RayCluster controller")
 }
 
 // +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;update
--- a/.github/workflows/olm_tests.yaml
+++ b/.github/workflows/olm_tests.yaml
@@ -72,6 +72,13 @@ jobs:
           kubectl create namespace openshift-operators
           kubectl create -f .github/resources-olm-upgrade/operatorgroup.yaml
 
+      - name: Install Required KubeRay CRDs for Pod Initialisation
+        run: |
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -17,6 +17,16 @@ RUN CGO_ENABLED=1 GOOS=linux GOARCH=amd64 go build -tags strictfipsruntime -a -o
 
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.8
 WORKDIR /
+
+RUN microdnf update -y && \
+    microdnf install tar -y \
+    && microdnf clean all && \
+    rm -rf /var/cache/yum
+ADD https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable/openshift-client-linux.tar.gz /tmp/openshift-client-linux.tar.gz
```
dimakis, line 25: use this instead please https://gitlab.cee.redhat.com/data-hub/rhods-cpaas-midstream/-/blob/rhods-1.34-rhel-8/distgit/containers/odh-deployer/Dockerfile.in?ref_type=heads

config/manager/manager.yaml
```diff
@@ -31,6 +31,28 @@ spec:
...
+        - sh
+        - -c
+        - |
+          set -e
+          CRDS="rayclusters.ray.io rayjobs.ray.io rayservices.ray.io"
+          for crd in $CRDS; do
+            echo "Checking for $crd"
+            until oc get crd $crd; do
+              echo "$crd not available yet, retrying in 10 seconds..."
+              sleep 10
```
zdtsw, line 52: what happens CRD is not installed, not just not available for short period? 
this will cause initContainer never stop?

zdtsw:
i must have missed something again:
you are injecting "oc" (or kubectl) binary into the same CFO operator image, and start an initContainer with exactly the same image, for one purpose to check if these 3 CRD has been installed into the cluster?


But, why not:
- find a dedicated initcontainer images as initContainer (e.g the one dimakis pointed out?) all you need is just the "oc" binary
- do not touch existing Dockerfile.
- and no need use var in kustomize but hardcode image with digists in the deployment of the CFO

ChristianZaccaria:
Hi @zdtsw, completely agreed, that makes more sense to me. Will fix now, thanks for the advice and insights!

config/manager/manager.yaml
ChristianZaccaria: That would be the current behaviour.  The CFO shouldn't start without those CRDs previously installed, hence, the initContainer will continue to search for them until they are present.

Should I have a timeout for it? In case of timing out, I suppose the only way of 're-activating' the CFO would be to restart the pod.

zdtsw: so if the bash script timeout, it should exit non-zero, which by default will cause Pod restart without start the "real" container.
i think this can be easily verified in the cluster.

Dockerfile
ChristianZaccaria: I'm trying to pull `registry.redhat.io/openshift4/ose-cli:v4.13` but seem to get a permission error from registry.redhat.io. Looking into it.

astefanutti (commented):
Is this solution compatible with the webhooks that are being introduced in #507 and #508?

astefanutti (commented):
An alternative approach would be to setup a watcher in the CFO main if the CRDs are not present, that would watch their installation, and start the controller and webhooks on that event.

main.go
```diff
@@ -127,6 +130,15 @@ func main() {
 	kubeClient, err := kubernetes.NewForConfig(kubeConfig)
 	exitOnError(err, "unable to create Kubernetes client")
 
+	apiextensionsClient, err := clientset.NewForConfig(kubeConfig)
+	if err != nil {
+		exitOnError(err, "unable to create apiextensionsClient")
+	}
+	crdName := "rayclusters.ray.io"
+	if err := checkCRDAvailability(apiextensionsClient, crdName); err != nil {
```
astefanutti, line 138: How is it different from the call to `hasAPIResourceForGVK` that's done a bit after.

ChristianZaccaria: TBH I only noticed that new function just now after rebasing. - The main difference is that with the Kubernetes API Extensions client we can directly check for the required CRD, which is more focused on the scope of this issue.

With that said, running the CFO from main branch with the `hasAPIResourceForGVK` approach, for some reason, without the rayclusters CRD installed it doesn't fail the initialisation of the CFO pod.

astefanutti: Functionally these two methods does very much the same. It doesn't fail on main only because we want it, we just do not start the RayCluster controller at the moment

`hasAPIResoureForGVK` has the advantage that it doesn't require permission to read CRDs.

ChristianZaccaria: That's true, a good advantage. One question in case I'm mistaken, do we really want to start the CFO anyways without the RC Controller? Even after applying the CRD, the CFO won't attempt to start the controller. We would remain in the same scenario where the user would need to restart the pod. Moreover, I couldn't find any error messages displayed in the CFO logs for it.

astefanutti: The idea was that there may be other controllers orthogonal to KubeRay, like in #491. But it's true at the moment we could exit in that case.

ChristianZaccaria: I see, thanks for clarifying on that, I wasn't sure. I refactored the logic to exit and not initialise the pod unless the CRD is present. Let me know what you think. I know this is a temporary change too until the RC Controller is moved to KubeRay.

ChristianZaccaria: Current behaviour now: if the RayCluster CRD is not available, the CFO will fail to start and attempt to restart until the CRD is installed - This would display error logs in the CFO pod. Once the RayCluster CRD is installed, the CFO manager and RC Controller will start

ChristianZaccaria:
Closing in favour of https://github.com/project-codeflare/codeflare-operator/pull/546

ChristianZaccaria:
Closing in favour of https://github.com/project-codeflare/codeflare-operator/pull/546

PR #513, opened 2024-04-12
Removed make install & make uninstall target

Ygnas:
# Issue link

[RHOAIENG-5347](https://issues.redhat.com/browse/RHOAIENG-5347)

# What changes have been made
Removed make targets 

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [x] Testing is not required for this change

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,6 @@ For building and pushing a new version of the operator image:
 
 For deploying onto a cluster:
  - First, either set `KUBECONFIG` or ensure you are logged into a cluster in your environment
- - `make install`
  - `make deploy -e IMG=<image-repo/image-name>`
 
 For building and pushing a new version of the bundled operator image:
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Requirements:
   When you have a version of the GNU sed installed on a macOS you may specify the binary using
   ```bash
   # brew install gnu-sed
-  make install -e SED=/usr/local/bin/gsed
+  NAMESPACE=default make run
   ```
 
 ### Testing
@@ -34,8 +34,6 @@ The e2e tests can be executed locally by running the following commands:
     ```bash
     # Create a KinD cluster
     make kind-e2e
-    # Install the CRDs
-    make install
     ```
 
    [!NOTE]
--- a/Makefile
+++ b/Makefile
@@ -173,16 +173,6 @@ ifndef ignore-not-found
   ignore-not-found = false
 endif
 
-.PHONY: install
-install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/crd | kubectl apply -f -
-	git restore config/*
-
-.PHONY: uninstall
-uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
-	$(KUSTOMIZE) build config/crd | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
-	git restore config/*
-
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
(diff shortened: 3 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -34,10 +34,9 @@ For building and pushing a new version of the operator image:
  - `make image-build -e IMAGE_TAG_BASE=<image-repo/image-name> VERSION=<semver>`
  - `make image-push -e IMAGE_TAG_BASE=<image-repo/image-name> VERSION=<semver>`
 
-For deploying onto a cluster:
+For running from your host:
  - First, either set `KUBECONFIG` or ensure you are logged into a cluster in your environment
- - `make install`
- - `make deploy -e IMG=<image-repo/image-name>`
+ - `make run`
```
sutaakar, line 39: AFAIK you need to provide namespace:
`NAMESPACE=default make run`

README.md
```diff
@@ -22,7 +22,7 @@ Requirements:
   When you have a version of the GNU sed installed on a macOS you may specify the binary using
   ```bash
   # brew install gnu-sed
-  make install -e SED=/usr/local/bin/gsed
+  NAMESPACE=default make run -e SED=/usr/local/bin/gsed
```
zdtsw, line 25: why should this "make install" eventually become "make run" ?

CONTRIBUTING.md
```diff
@@ -34,10 +34,9 @@ For building and pushing a new version of the operator image:
  - `make image-build -e IMAGE_TAG_BASE=<image-repo/image-name> VERSION=<semver>`
  - `make image-push -e IMAGE_TAG_BASE=<image-repo/image-name> VERSION=<semver>`
 
-For deploying onto a cluster:
+For running from your host:
  - First, either set `KUBECONFIG` or ensure you are logged into a cluster in your environment
- - `make install`
- - `make deploy -e IMG=<image-repo/image-name>`
+ - `NAMESPACE=default make run`
```
zdtsw, line 39: why should use "default" ns here, and "make install deploy" becomes "go run"?

Ygnas: You are right here.

README.md
Ygnas: Make install does not work at the moment, so people following the readme with get an error. Is changing it to make run bad here?

sutaakar: I would say that previous instructions were wrong, as `make install` just installed CRDs, however it didn't start the operator itself.
`make run` will start the operator now, connecting to the cluster from Kubeconfig.

sutaakar: @Ygnas Can you check if `-e SED=/usr/local/bin/gsed` is still needed. AFAIK there is no string replacement being done as part of `make run`

Ygnas: @sutaakar Yes `make run` does not need sed to run and it works fine. So would leaving only a note in this place fine?

zdtsw: if you want use "make run" here, just remove the "sed" part

Btw: @Fiona-Waters  + @dimakis looks like this PR: 
 https://github.com/project-codeflare/codeflare-operator/commit/4b863f79d4b704c511ea404e27e3ec65d66b5a0f removed the support of gsed on MacOS, was that a mistake?   FYI @tedhtchang

sutaakar: @Ygnas thinking, is SED used in any other command? If not then you can remove the note too.

sutaakar: ok, so just remove the SED from command

sutaakar (commented):
/lgtm

README.md
Fiona-Waters: > if you want use "make run" here, just remove the "sed" part
> 
> Btw: @Fiona-Waters + @dimakis looks like this PR: [4b863f7](https://github.com/project-codeflare/codeflare-operator/commit/4b863f79d4b704c511ea404e27e3ec65d66b5a0f) removed the support of gsed on MacOS, was that a mistake? FYI @tedhtchang

@zdtsw This was completed quickly just before a release as far as I remember. It was updated very soon afterwards in this [commit](https://github.com/project-codeflare/codeflare-operator/commit/c7954a9bd6b19f5978bcf318ebb7d839ad4b1b99) - a lot of this has been removed now with the removal of mcad. Hope I am understanding correctly.

dgrove-oss:
We're going to be adding CRDs back again next week (when we bring in the AppWrapper operator).

Ygnas:
> We're going to be adding CRDs back again next week (when we bring in the AppWrapper controller).

Closing because of this.

PR #526, opened 2024-04-16
add watches for owned resources and CRBs

KPostOffice:
# Issue link

# What changes have been made
I've added watches for owned resources so that the controller will automatically re-reconcile on updates to owned resources. I've also added a label to the CRB make mapping from CRB to RayCluster resource possible

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/pkg/controllers/raycluster_controller.go
+++ b/pkg/controllers/raycluster_controller.go
@@ -34,6 +34,7 @@ import (
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
 
 	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -48,6 +49,8 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	routev1 "github.com/openshift/api/route/v1"
 	routev1ac "github.com/openshift/client-go/route/applyconfigurations/route/v1"
@@ -78,6 +81,8 @@ const (
 
 	CAPrivateKeyKey = "ca.key"
 	CACertKey       = "ca.crt"
+
+	RayClusterNameLabel = "ray.openshift.ai/cluster-name"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


KPostOffice:
/hold
Need to test

pkg/controllers/raycluster_controller.go
```diff
@@ -220,7 +222,7 @@ func crbNameFromCluster(cluster *rayv1.RayCluster) string {
 func desiredOAuthClusterRoleBinding(cluster *rayv1.RayCluster) *rbacapply.ClusterRoleBindingApplyConfiguration {
 	return rbacapply.ClusterRoleBinding(
```
astefanutti, line 223: I really wonder, can this be a RoleBinding instead of a ClusterRoleBinding?

KPostOffice: I believe I tried this a while back and it didn't work. I can give it another go though

pkg/controllers/raycluster_controller.go
```diff
@@ -233,7 +237,7 @@ func crbNameFromCluster(cluster *rayv1.RayCluster) string {
...
 		crbNameFromCluster(cluster)).
-		WithLabels(map[string]string{"ray.io/cluster-name": cluster.Name}).
+		WithLabels(map[string]string{"ray.io/cluster-name": cluster.Name, "ray.io/cluster-namespace": cluster.Namespace}).
```
astefanutti, line 240: Can we scope the labels to `ray.openshift.ai` instead of `ray.io` that we do not own?

pkg/controllers/raycluster_controller.go
```diff
@@ -350,5 +354,19 @@ func (r *RayClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		Owns(&corev1.ServiceAccount{}).
+		Owns(&corev1.Service{}).
+		Owns(&corev1.Secret{}).
+		Owns(&routev1.Route{}).
+		Owns(&networkingv1.Ingress{}).
+		Watches(&rbacv1.ClusterRoleBinding{}, handler.EnqueueRequestsFromMapFunc(
+			func(c context.Context, o client.Object) []reconcile.Request {
+				return []reconcile.Request{{
+					NamespacedName: client.ObjectKey{
+						Name:      o.GetLabels()["ray.io/cluster-name"],
```
astefanutti, line 366: Should we check for the absence of these labels? The watch will receive events for all CRBs. An alternative is to configure the cache to filter events only for our CRBs.

KPostOffice: I've changed it to check for the absence of the labels, I'm not sure how to configure the cache to filter for only our CRBs

astefanutti: I've created https://github.com/project-codeflare/codeflare-operator/tree/pr-cache-example that shows how to configure the manager cache.

KPostOffice: Thanks Antonin, I'm adding this too so that it can be followed if someone finds this
https://github.com/kubernetes-sigs/controller-runtime/blob/main/designs/use-selectors-at-cache.md

pkg/controllers/support.go
```diff
@@ -20,7 +20,7 @@ func serviceNameFromCluster(cluster *rayv1.RayCluster) string {
 
 func desiredRayClientRoute(cluster *rayv1.RayCluster) *routeapply.RouteApplyConfiguration {
 	return routeapply.Route(rayClientNameFromCluster(cluster), cluster.Namespace).
-		WithLabels(map[string]string{"ray.io/cluster-name": cluster.Name}).
+		WithLabels(map[string]string{"ray.openshift.ai/cluster-name": cluster.Name}).
```
astefanutti, line 23: Should that be `RayClusterNameLabel`?

astefanutti (commented):
Just left  a comment, otherwise LGTM. Great work!

pkg/controllers/support.go
KPostOffice: Yep, did a find and replace in the controller code. Missed this one. Thanks for catching

KPostOffice: I also added WithController to the OwnerReferences in support.go

dgrove-oss: For the record, this broke the AppWrapper controller.   I submitted a fix here #640

pkg/controllers/raycluster_controller.go
dgrove-oss: For the record, this broke the AppWrapper controller.   I submitted a fix here #640

PR #527, opened 2024-04-17
Remove owner reference for ClusterRoleBinding to RayCluster

astefanutti:
An owner reference from a non-namespaced resource cannot be namespaced.


pkg/controllers/raycluster_controller.go
```diff
@@ -232,9 +232,6 @@ func desiredOAuthClusterRoleBinding(cluster *rayv1.RayCluster) *rbacapply.Cluste
 				WithAPIGroup("rbac.authorization.k8s.io").
 				WithKind("ClusterRole").
 				WithName("system:auth-delegator"),
-		).
-		WithOwnerReferences(
-			v1.OwnerReference().WithUID(cluster.UID).WithName(cluster.Name).WithKind(cluster.Kind).WithAPIVersion(cluster.APIVersion),
```
sutaakar, line 237: In such case controller needs to make sure the CRB is deleted when RayCluster is removed.

KPostOffice (approved):

PR #522, opened 2024-04-16
add: mutating webhook

VanillaSpoon:
# Issue link
https://issues.redhat.com/browse/RHOAIENG-1991

# What changes have been made
Add Mutating Webhook to add OAuth SideCar to RayCluster

# Verification steps
Setup environment.
Build and Deploy the codeflare-operator

Ensure all configurations have been applied

From the sdk, create a raycluster
- Ensure the raycluster is deployed correctly
- Inspect the raycluster yaml to enusre the oauth configuraitons have been applied
- Inspect the raycluster head node and ensure the oauth is running alongside it.

## Checks
- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [ ] Testing is not required for this change

PR #530, opened 2024-04-17
Add RayCluster mutating webhook

astefanutti:
Includes #522


config/webhook/service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: webhook-service
+  namespace: openshift-operators
```
sutaakar, line 5:
```suggestion
  namespace: system
```

pkg/controllers/raycluster_webhook.go
```diff
@@ -0,0 +1,134 @@
...
+	for _, container := range raycluster.Spec.HeadGroupSpec.Template.Spec.Containers {
+		if container.Name == "oauth-proxy" {
+			rayclusterlog.V(2).Info("OAuth sidecar already exists, no patch needed")
+			return nil
+		}
+	}
+
+	rayclusterlog.V(2).Info("Adding OAuth sidecar container")
+	// definition of the new container
+	newOAuthSidecar := corev1.Container{
```
sutaakar, line 72: Would it have sense to limit Resources for oauth container?

config/openshift/kustomization.yaml
```diff
@@ -0,0 +1,25 @@
...
+commonLabels:
+  app.kubernetes.io/name: codeflare-operator
+  app.kubernetes.io/part-of: codeflare
+
+bases:
+  - ../default
+  - ../webhook
+
+resources:
+  - metrics_service.yaml
```
sutaakar, line 21: This should be removed

astefanutti: Good catch, removed.

sutaakar (commented):
/lgtm

pkg/controllers/raycluster_webhook.go
astefanutti: Right, for the sake of keeping that PR from growing too big, I'd suggest to add that in a separate PR.

Bobbins228 (commented):
/lgtm

PR #533, opened 2024-04-18
Use cert-controller to generate webhook certificates

astefanutti:


main.go
```diff
@@ -127,34 +130,52 @@ func main() {
...
 
 	OpenShift := isOpenShift(ctx, kubeClient.DiscoveryClient)
 
-	if OpenShift {
-		// TODO: setup the RayCluster webhook on vanilla Kubernetes
-		exitOnError(controllers.SetupRayClusterWebhookWithManager(mgr, cfg.KubeRay), "error setting up RayCluster webhook")
-	}
+	go setupControllers(mgr, kubeClient, cfg, OpenShift, certsReady)
+
+	exitOnError(mgr.AddHealthzCheck(cfg.Health.LivenessEndpointName, healthz.Ping), "unable to set up health check")
```
dgrove-oss, line 164: Do you want to grab the code block from the AppWrapper controller that defers the health/readiness probes until the certificates or ready?  I called it in #491, but probably best to copy it to the codeflare operator's main.  Code is here: https://github.com/project-codeflare/appwrapper/blob/2213dec59688ddc2259f2d0c5cea4ad759b0e047/pkg/controller/setup.go#L96-L112

astefanutti: Right, that's a good idea. Added in af8f912f2218f3a1356a880dbc44af8c592950e1.

sutaakar (commented):
/lgtm

dimakis (approved):

PR #535, opened 2024-04-18
Remove dangling MCAD and InstaScale references

dgrove-oss:
Cleanup a few MCAD/InstaScale references missed in c771aae4.


sutaakar (commented):
/lgtm

PR #547, opened 2024-04-30
Allow for custom CertGeneratorImage

Fiona-Waters:
# Issue link
[RHOAIENG-6505](https://issues.redhat.com/browse/RHOAIENG-6505)

# What changes have been made
I updated the CFO config to include an image to be used for cert creation in the init containers added by the CFO webhook. This is added so that in the case of a disconnected cluster there is not an extra image required.

# Verification steps
- Install CodeFlare Operator from this branch, and provide a custom image in the `codeflare-operator-config` configMap adding the image to the kuberay configuration like this:
```
    kuberay:
      ingressDomain: ""
      mTLSEnabled: true
      rayDashboardOAuthEnabled: true
      certGeneratorImage: "registry.access.redhat.com/ubi9@sha256:770cf07083e1c85ae69c25181a205b7cdef63c11b794c89b3b487d4670b4c328"
```
- Restart the CFO pod.
- Create a raycluster using a demo notebook, and check that the init containers are using the image specified in the configmap and that the raycluster pods are created succesfully. 

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [X] Manual tests

### DIFF
--- a/config/e2e/config.yaml
+++ b/config/e2e/config.yaml
@@ -7,3 +7,4 @@ data:
     kuberay:
       rayDashboardOAuthEnabled: false
       ingressDomain: "kind"
+      certGeneratorImage: "quay.io/project-codeflare/ray:latest-py39-cu118"
--- a/config/e2e/config.yaml
+++ b/config/e2e/config.yaml
@@ -7,4 +7,4 @@ data:
     kuberay:
       rayDashboardOAuthEnabled: false
       ingressDomain: "kind"
-      certGeneratorImage: "quay.io/project-codeflare/ray:latest-py39-cu118"
+      mTLSEnabled: false
--- a/main.go
+++ b/main.go
@@ -129,6 +129,7 @@ func main() {
 			RayDashboardOAuthEnabled: ptr.To(true),
 			IngressDomain:            "",
 			MTLSEnabled:              ptr.To(true),
+			CertGeneratorImage:       "quay.io/project-codeflare/ray:latest-py39-cu118",
 		},
 	}
 
--- a/main.go
+++ b/main.go
@@ -129,7 +129,7 @@ func main() {
 			RayDashboardOAuthEnabled: ptr.To(true),
 			IngressDomain:            "",
 			MTLSEnabled:              ptr.To(true),
-			CertGeneratorImage:       "quay.io/project-codeflare/ray:latest-py39-cu118",
+			CertGeneratorImage:       "registry.access.redhat.com/ubi9@sha256:770cf07083e1c85ae69c25181a205b7cdef63c11b794c89b3b487d4670b4c328",
 		},
 	}
 
--- a/pkg/controllers/raycluster_webhook.go
+++ b/pkg/controllers/raycluster_webhook.go
@@ -87,7 +87,7 @@ func (w *rayClusterWebhook) Default(ctx context.Context, obj runtime.Object) err
 		}
 
 		// Append the create-cert Init Container
-		rayCluster.Spec.HeadGroupSpec.Template.Spec.InitContainers = upsert(rayCluster.Spec.HeadGroupSpec.Template.Spec.InitContainers, rayHeadInitContainer(rayCluster, w.Config.IngressDomain, w.Config.CertGeneratorImage), withContainerName(initContainerName))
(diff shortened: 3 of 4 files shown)


pkg/controllers/raycluster_webhook.go
```diff
@@ -161,8 +161,8 @@ func (w *rayClusterWebhook) ValidateUpdate(ctx context.Context, oldObj, newObj r
 
 	// Init Container related errors
 	if ptr.Deref(w.Config.MTLSEnabled, true) {
-		allErrors = append(allErrors, validateHeadInitContainer(rayCluster, w.Config.IngressDomain)...)
-		allErrors = append(allErrors, validateWorkerInitContainer(rayCluster)...)
+		allErrors = append(allErrors, validateHeadInitContainer(rayCluster, w.Config.IngressDomain, w.Config.CertGeneratorImage)...)
+		allErrors = append(allErrors, validateWorkerInitContainer(rayCluster, w.Config.CertGeneratorImage)...)
```
KPostOffice, line 165: Can we change these to either take `w` or `w.Config` directly that way if in the future other parts of the struct are needed they can be dereferenced directly in the function rather than adding more and more args.

https://refactoring.guru/preserve-whole-object

main.go
```diff
@@ -129,6 +129,7 @@ func main() {
 			RayDashboardOAuthEnabled: ptr.To(true),
 			IngressDomain:            "",
 			MTLSEnabled:              ptr.To(true),
+			CertGeneratorImage:       "quay.io/project-codeflare/ray:latest-py39-cu118",
```
KPostOffice, line 132: I would like to change this to UBI9 pending validating that the init container successfully executes

KPostOffice (commented):
Looks good, just a couple comments 👍🏻

ChughShilpa:
@Fiona-Waters 
One question, To use custom image, user needs to specify it in cluster config in demo-notebooks and then also need to update the same in `codeflare-operator-config` config-map ?

Fiona-Waters:
> @Fiona-Waters One question, To use custom image, user needs to specify it in cluster config in demo-notebooks and then also need to update the same in `codeflare-operator-config` config-map ?

The ray image specified in the cluster config in demo-notebooks will be the main ray image used. The one set in the config map will be used just for tls cert creation in the init containers. It can be the same image for both but it doesn't have to be.

pkg/controllers/raycluster_webhook.go
Fiona-Waters: Updated to take w.config instead. Thank you.

main.go
Fiona-Waters: I have updated this to `registry.access.redhat.com/ubi9@sha256:770cf07083e1c85ae69c25181a205b7cdef63c11b794c89b3b487d4670b4c328` and the raycluster pods come up as expected with init containers running successfully with this image. 
I don't think it's related as I am getting the same error when using the other default image but when running the basic_interactive demo notebook I'm seeing `RuntimeError: Python minor versions differ between client and server: client is 3.9.16, server is 3.8.13` when runninng the cell containing ray.init.

Fiona-Waters: Also the e2e is failing, looking into it.

Fiona-Waters: When running it manually on openshift I'm seeing this error in the head pod `Permission denied: '/proc/driver/nvidia/gpus'`

Fiona-Waters: Needed to set mtls to false in e2e config. Working successfully with UBI9 image.

config/e2e/config.yaml
```diff
@@ -7,3 +7,4 @@ data:
     kuberay:
       rayDashboardOAuthEnabled: false
       ingressDomain: "kind"
+      mtLSEnabled: false
```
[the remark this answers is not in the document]
Fiona-Waters: Updated. It is needed as the e2e test is out of date. Creating a follow on jira to cover this.

KPostOffice (approved):
/lgtm

config/e2e/config.yaml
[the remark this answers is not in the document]
Fiona-Waters: https://issues.redhat.com/browse/RHOAIENG-7055

PR #549, opened 2024-05-08
Check if workerGroupSpec is not empty before accessing it

ChristianZaccaria:
# Issue link

Jira: https://issues.redhat.com/browse/RHOAIENG-6614 

# What changes have been made

- The failing Ray Tests were caused by an `index out of range` error in the webhook logic. The `workerGroupSpec` was attempted to be accessed when it doesn't exist.
- Check is now made to verify it's not empty before accessing it.

# Verification steps

0. Have all components deployed including this version of the CFO and Kueue.
1. Deploy KubeRay-Operator in the same namespace as referenced in the [DSCI `default-dsci`](https://github.com/project-codeflare/codeflare-operator/blob/a7619510bca4d70ed4362846703735f027106f10/pkg/controllers/raycluster_controller.go#L254-L256)
2. In the KubeRay repo, run the previously failing test:
`go test -timeout 30m -v ./test/e2e/rayjob_lightweight_test.go ./test/e2e/support.go`
3. Test should pass.

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


pkg/controllers/raycluster_webhook.go
```diff
@@ -100,24 +100,25 @@ func (w *rayClusterWebhook) Default(ctx context.Context, obj runtime.Object) err
...
-			rayCluster.Spec.WorkerGroupSpecs[0].Template.Spec.Volumes = upsert(rayCluster.Spec.WorkerGroupSpecs[0].Template.Spec.Volumes, caVol, withVolumeName(caVol.Name))
-		}
-
-		// Append the certificate volume mounts
-		for _, mount := range certVolumeMounts() {
-			rayCluster.Spec.WorkerGroupSpecs[0].Template.Spec.Containers[0].VolumeMounts = upsert(rayCluster.Spec.WorkerGroupSpecs[0].Template.Spec.Containers[0].VolumeMounts, mount, byVolumeMountName)
+		if len(rayCluster.Spec.WorkerGroupSpecs) != 0 {
+			// Append the list of environment variables for the worker container
+			for _, envVar := range envVarList() {
+				rayCluster.Spec.WorkerGroupSpecs[0].Template.Spec.Containers[0].Env = upsert(rayCluster.Spec.WorkerGroupSpecs[0].Template.Spec.Containers[0].Env, envVar, withEnvVarName(envVar.Name))
```
sutaakar, line 106: RayCluster can contain more WorkerGroupSpecs. Would it have sense to rather loop all existing WorkerGroupSpecs to verify/modify all of them?

sutaakar:
@ChristianZaccaria Can you please provide a unit test verifying the changes?

pkg/controllers/raycluster_webhook.go
ChristianZaccaria: That makes complete sense, great catch thanks Karel!

ChristianZaccaria:
@sutaakar I added unit tests covering these changes. Please let me know if there are other test cases I should be covering here. 

Also, I am thinking of adding a RayCluster template in the test support file and perhaps append items to it in the tests to reduce the amount of code in the test file. Is this a good approach or what is the best practices on this? Thank you! :)

pkg/controllers/raycluster_webhook_test.go
```diff
@@ -0,0 +1,615 @@
...
+	test.Client().Ray().RayV1().RayClusters(namespace).Create(test.Ctx(), invalidRayCluster, metav1.CreateOptions{})
+
+	// Positive test case (expect no errors)
+	validWorkerInitContainer := validateWorkerInitContainer(validRayCluster)
+	test.Expect(validWorkerInitContainer).To(gomega.BeEmpty())
+
+	// Negative test case (expect errors for missing Init Container)
+	invalidWorkerInitContainer := validateWorkerInitContainer(invalidRayCluster)
+	test.Expect(invalidWorkerInitContainer).NotTo(gomega.BeEmpty())
+	test.Expect(invalidWorkerInitContainer[0].Detail).To(gomega.ContainSubstring("create-cert Init Container is immutable"))
```
sutaakar, line 176: One general comment - can you split test cases, so every test case is validated separately?
i.e. here you validate both valid and invalid case, so in case positive test case fails then negative test case is not validated.

sutaakar:
@ChristianZaccaria Creating one valid RayCluster CR and then modify it is a possible approach, may increase test readability.

pkg/controllers/raycluster_webhook_test_support.go
```diff
@@ -0,0 +1,43 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+
+func mutateWorkerGroupSpecs(rayCluster *rayv1.RayCluster) {
```
sutaakar, line 21: Why did you move this function into separate file?

ChristianZaccaria: The changes in this PR affect parts of the webhook [Default function](https://github.com/project-codeflare/codeflare-operator/blob/main/pkg/controllers/raycluster_webhook.go#L68-L124). I thought for testing that specific block of code, I would need to separate it and call that function on its own. The other approach could be to leave it where it was and assert an exact output or assert the length of items perhaps.

sutaakar: Thinking whether it would have sense to verify result of the "main" function instead of checking every utility function....
i.e. call `Default(...)` or `ValidateCreate(...)` and so on.

sutaakar: That approach will make tests independent on internal implementation functions in webhook, also it will verify whole workflow for the test cases.

ChristianZaccaria: I think I get what you mean. So instead of calling the utility functions in the tests, I should call the "main" functions which will call those utility functions too. This way, basically testing the expected output from the functionality of the whole file right? - Sounds good to me, thanks for advising!

pkg/controllers/raycluster_webhook_test.go
ChristianZaccaria: Ohh that's right! Thanks - will split test cases.

pkg/controllers/raycluster_webhook_test_support.go
sutaakar: right

pkg/controllers/raycluster_webhook_test.go
```diff
@@ -0,0 +1,393 @@
...
+	// Create the RayClusters
+	test.Client().Ray().RayV1().RayClusters(namespace).Create(test.Ctx(), validRayCluster, metav1.CreateOptions{})
+
+	// Call to ValidateCreate function is made
+	warnings, err := rcWebhook.ValidateCreate(test.Ctx(), runtime.Object(validRayCluster))
+	t.Run("Expected no warnings or errors on call to ValidateCreate function", func(t *testing.T) {
+		g := gomega.NewWithT(t)
+		g.Expect(warnings).To(gomega.BeNil())
+		g.Expect(err).To(gomega.BeNil())
+	})
```
ChristianZaccaria, line 374: @sutaakar I only then realised that this basically covers all test cases. Perhaps the tests after this one are not needed?

Also, If I do the same for the Default function, would all the other tests/assertions not be required?

Open to thoughts on this before continuing, thanks!

ChristianZaccaria: Thinking perhaps the assertions made for the Default (mutating) function is required to compare outputs. But, for the validating functions such as `ValidateCreate`, as long as no errors are returned (meaning all elements exist in the defined RayCluster), the validation should succeed.

sutaakar: For `ValidateCreate` it is enough to have one positive test, providing negative tests to validate incorrect modification and its detection.

pkg/controllers/raycluster_webhook_test.go
```diff
@@ -0,0 +1,397 @@
...
+		g := gomega.NewWithT(t)
+		g.Expect(func() bool {
+			for _, container := range validRayCluster.Spec.HeadGroupSpec.Template.Spec.Containers {
+				if container.Name == oauthProxyContainerName {
+					return true
+				}
+			}
+			return false
+		}()).To(gomega.BeTrue(), "Expected the OAuth proxy container to be present in the list of head group containers")
+	})
```
sutaakar, line 111: This boilerplate isn't needed, you can leverage `test` directly as it embeds gomega:

```
	test.Expect(err).NotTo(HaveOccurred(), "Expected no errors on call to Default function")
	test.Expect(validRayCluster.Spec.HeadGroupSpec.Template.Spec.Containers).
		To(ContainElement(WithTransform(ContainerName, Equal(oauthProxyContainerName))), "Expected the OAuth proxy container to be present in the list of head group containers")
```

With helper functions in the bottom (to be relocated into codeflare-common):
```
func ContainerName(container corev1.Container) string {
	return container.Name
}
```

sutaakar: Also this approach avoids `BeTrue()` checks for element with specific value, which often doesn't provide meaningful context.

sutaakar: Test can also check number of elements, making the assertion more structured:
```
	test.Expect(validRayCluster.Spec.HeadGroupSpec.Template.Spec.Containers).
		To(
			And(
				HaveLen(1),
				ContainElement(WithTransform(ContainerName, Equal(oauthProxyContainerName))),
			),
			"Expected the OAuth proxy container to be present in the list of head group containers",
		)
```

pkg/controllers/raycluster_webhook_test.go
sutaakar: Or possibly to have several tests, providing valid RayClusters, i.e. empty workerGroupSpecs and such.

ChristianZaccaria: Hopefully, latest changes in this PR reflects the above. - One positive test case, providing negative tests for manipulated fields. Thanks!

ChristianZaccaria:
@sutaakar latest changes include one positive test for each "main" function, and several negative tests to cover the full functionality of the raycluster_webhook.go file.

pkg/controllers/raycluster_webhook_test.go
```diff
@@ -0,0 +1,668 @@
...
+		test.Expect(warnings).Should(BeNil(), "Expected no warnings on call to ValidateUpdate function")
+		test.Expect(err).ShouldNot(HaveOccurred(), "Expected no errors on call to ValidateUpdate function")
+	})
+
+	// Negative Test Cases
+	trueBool := true
+	invalidRayCluster := validRayCluster.DeepCopy()
+
+	t.Run("Negative: Expected errors on call to ValidateUpdate function due to EnableIngress set to True", func(t *testing.T) {
+		invalidRayCluster.Spec.HeadGroupSpec.EnableIngress = &trueBool
```
sutaakar, line 544: Nitpick, there is a dedicate funtion to create pointer for values:
```suggestion
		invalidRayCluster.Spec.HeadGroupSpec.EnableIngress = support.Ptr(true)
```

sutaakar:
@ChristianZaccaria Can you also adjust `test-unit` Makefile target to run the new unit test (currently it doesn't run anything). That way the unit test will be executed for PRs.

Otherwise this PR looks great, good job.

ChristianZaccaria:
@sutaakar I made the last few changes, tests are now running as part of this PR check in the `Unit tests` workflow. Thank you!

ChristianZaccaria:
I made a [PR to move the core functions](https://github.com/project-codeflare/codeflare-common/pull/48) used here to the codeflare-common repository. Should we merge that and then make the changes here or what is the appropriate approach? Thanks @sutaakar

sutaakar:
/lgtm

> I made a https://github.com/project-codeflare/codeflare-common/pull/48 used here to the codeflare-common repository. Should we merge that and then make the changes here or what is the appropriate approach?

That is up to you, can be merged now and refactored later or it can wait.

PR #558, opened 2024-05-21
upgrade AppWrapper from v0.11.1 to v0.12.0

dgrove-oss:


dgrove-oss:
@astefanutti -- v0.12.0 of AppWrapper includes a fix to https://github.com/project-codeflare/appwrapper/issues/142 to remove the dependency on k8s.io/kubernetes

PR #574, opened 2024-06-19
Mutate and Validate RayCluster on SecurityContext

ChristianZaccaria:
# Issue link

Jira: https://issues.redhat.com/browse/RHOAIENG-7638 

# What changes have been made

- The Mutating Webhook will now add the required `SecurityContext` in the RayCluster's head and worker containers.
- The Validating Webhook will now check for the required `SecurityContext` in the RayCluster's head and worker containers.

# Verification steps

1. Create Data science project in RHOAI
2. Create and start workbench
3. Find OpenShift namespace corresponding to the Data science project name, add labels to the `namespace` resource with value:
- pod-security.kubernetes.io/enforce: restricted
- pod-security.kubernetes.io/enforce-version: v1.24
4. Clone SDK demo notebooks in Workbench
5. Setup Kueue resources
6. Run i.e. 0_basic_ray.ipynb Notebook, try to create Ray cluster
7. RayCluster should start with or without the labels set in the `Namespace` resource.
8. You may also check that the `SecurityContext` is added to both head and worker containers when the mentioned labels have been set.
9. Attempt to create a RayCluster without the `SecurityContext`, and attempt to update the `SecurityContext` once the RayCluster is created. The ValidatingWebhook should reject these actions.

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/pkg/controllers/raycluster_webhook.go
+++ b/pkg/controllers/raycluster_webhook.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"reflect"
 	"strconv"
 
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
@@ -133,6 +134,7 @@ func (w *rayClusterWebhook) ValidateCreate(ctx context.Context, obj runtime.Obje
 	var allErrors field.ErrorList
 
 	allErrors = append(allErrors, validateIngress(rayCluster)...)
+	allErrors = append(allErrors, validateSecurityContext(rayCluster)...)
 
 	if ptr.Deref(w.Config.RayDashboardOAuthEnabled, true) {
 		allErrors = append(allErrors, validateOAuthProxyContainer(rayCluster)...)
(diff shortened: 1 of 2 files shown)


pkg/controllers/raycluster_webhook.go
```diff
@@ -123,6 +127,32 @@ func (w *rayClusterWebhook) Default(ctx context.Context, obj runtime.Object) err
 		}
 	}
 
+	hasSecurityLabels, err := namespaceHasSecurityLabels(ctx, w.Client, rayCluster.Namespace)
+	if err != nil {
+		rayclusterlog.Error(err, "Failed to check namespace resource labels")
+		return err
+	}
+	if hasSecurityLabels {
```
astefanutti, line 135: Why not adding the security context unconditionally?

ChristianZaccaria: I actually wasn't sure if that's what we wanted, hence made it conditional. On a security point a view, you're right, I think we should always have this securityContext enabled. Should I make it unconditional, and also add to the `ValidatingWebhook` to ensure these are set?

astefanutti: Yes, I think that should be added unconditionally. I'm not sure how it'll behave on non OpenShift clusters. We may make it only for OpenShift if it has to.

ChristianZaccaria: Hmm interesting point. I'll make the changes and check the e2e tests. This should be sufficient to confirm its behaviour. Thanks Antonin!

KPostOffice:
@ChristianZaccaria Should we close this then in lieu of https://github.com/opendatahub-io/kuberay/pull/19 ?

ChristianZaccaria:
Closing in lieu of https://github.com/opendatahub-io/kuberay/pull/19

PR #575, opened 2024-06-20
[RHOAIENG-9004] Adjust existing test and workflow for GPU testing

sutaakar:
# Issue link

[RHOAIENG-9004](https://issues.redhat.com/browse/RHOAIENG-9004)

# What changes have been made

Adjust GitHub e2e workflow to setup and install NVidia operator.
Convert Ray tests to use MNIST fashion test case on GPU. Adjust `TestMNISTPyTorchAppWrapper` to use GPU.
As a workaround for having just one GPU the tests are executed sequentially.

I also had to alter uploading CFO image to be loaded by KinD CLI.
The reason is that GPU image uses Docker and doesn't interact well with locally created registry.

Edit:
Thinking whether we should add GPU tests next to existing nonGPU ones, or rather replace them. Having both will increase number of tests, on the other side non GPU tests are faster and can be ran anywhere.

# Verification steps

Run e2e PR check - GPU is made available on KinD cluster and tests use it.

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/.github/workflows/e2e_tests.yaml
+++ b/.github/workflows/e2e_tests.yaml
@@ -27,7 +27,7 @@ concurrency:
 jobs:
   kubernetes-e2e:
 
-    runs-on: ubuntu-20.04-4core
+    runs-on: ubuntu-20.04-4core-gpu
 
     steps:
       - name: Checkout code
@@ -52,8 +52,14 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Setup NVidia GPU environment for KinD
+        uses: ./common/github-actions/nvidia-gpu-setup
+
       - name: Setup and start KinD cluster
         uses: ./common/github-actions/kind
+
+      - name: Install NVidia GPU operator for KinD
+        uses: ./common/github-actions/nvidia-gpu-operator
 
       - name: Deploy CodeFlare stack
         id: deploy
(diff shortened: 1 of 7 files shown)


test/e2e/mnist.py
```diff
@@ -176,13 +176,12 @@ def test_dataloader(self):
 
 # Initialize a trainer
 trainer = Trainer(
-    accelerator="auto",
+    accelerator="gpu",
```
astefanutti, line 179: Can that be made conditional?

test/e2e/mnist_fashion.py
```diff
@@ -0,0 +1,98 @@
...
+            optimizer.zero_grad()
+            pred = model(inputs)
+            loss = criterion(pred, labels)
+            loss.backward()
+            optimizer.step()
+        print(f"epoch: {epoch}, loss: {loss.item()}")
+
+
+# For GPU Training, set `use_gpu` to True.
+use_gpu = True
```
astefanutti, line 89: Can that be made conditional?

.github/workflows/e2e_tests.yaml
```diff
@@ -52,33 +52,41 @@ jobs:
...
       - name: Deploy CodeFlare stack
         id: deploy
         run: |
           echo Setting up CodeFlare stack
           make setup-e2e
 
           echo Deploying CodeFlare operator
-          IMG="${REGISTRY_ADDRESS}"/codeflare-operator
-          make image-push -e IMG="${IMG}"
+          IMG=localhost/codeflare-operator:0.0.1
```
astefanutti, line 71: Maybe `localhost/codeflare-operator` or `localhost/codeflare-operator:test`? `0.0.1` may be confusing.

Dockerfile
```diff
@@ -2,7 +2,7 @@
 
 # BEGIN -- workaround lack of go-toolset for golang 1.21
 
-ARG GOLANG_IMAGE=golang:1.21
+ARG GOLANG_IMAGE=docker.io/library/golang:1.21
```
astefanutti, line 5: Out-of-curiosity, why is that needed?

sutaakar: Without this change Podman tries to resolve registry for golang, which failed even though I copied configuration for registries in https://github.com/project-codeflare/codeflare-common/blob/fcf105adb00e1221585404498504ed6f5b53075a/github-actions/nvidia-gpu-setup/action.yml#L13

So I didn't want to dig too deeply into it and rather adjusted image to reference registry too.

test/e2e/mnist.py
sutaakar: Thinking if default value should be GPU or CPU

Dockerfile
astefanutti: Ah I see, thanks.

test/e2e/mnist.py
astefanutti: Maybe something like `'gpu' if torch.cuda.is_available() else 'cpu'`?

sutaakar: My intention is to have assurance of what accelerator is used.
So in case GPU isn't provisioned for some reason the GPU test should fail, to indicate that GPU wasn't tested. By using condition above the test will hide the fact that GPU wasn't tested.

sutaakar: By providing external configuration (i.e. env variable, or splitting tests to run both CPU and GPU variants) I can make sure what is tested.

astefanutti: I see, maybe an assertion somewhere else than the training script? Like in the test method?

astefanutti: Or an environment variable?

astefanutti: Ah we were typing at the same time and I didn't see you last suggestions :)

Yes, an environment variable looks like a good idea.

sutaakar:
@astefanutti Ready for second round of review.
Right now GH action runs both CPU and GPU tests. Thinking whether we should run only GPU tests on PR check, leaving CPU just for possible local execution (will speed up PR check).

test/e2e/mnist.py
sutaakar: Providing accelerator using env variable.

test/e2e/mnist_fashion.py
sutaakar: Refactored

astefanutti (approved):
Great!

Yes CPU is more for local development. It's OK to only run GPU on CI now.

Also if we are to refactor the MNIST test, I think it'd be good to remove PyTorch Lightning and use plain PyTorch, which should Lower the maintenance effort.

PR #521, opened 2024-04-16
fix(manifests): missing namespace in CM stack-config

zdtsw:
# Issue link

introduced in change https://github.com/opendatahub-io/codeflare-operator/commit/afe5bdc7fb34bddc72bbc8e177d31eaa9fb32bde
cause error:
`* field specified in var '{namespace ConfigMap.v1.[noGrp] {data.namespace}}' not found in corresponding resource`
which prevent deployment in ODH operator

# What changes have been made

# Verification steps

## Checks
- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [ ] Testing is not required for this change

PR #581, opened 2024-06-28
Update to AppWrapper v0.20.2

dgrove-oss:
@ChristianZaccaria @sutaakar  -- I believe this fixes the problem with a RayCluster wrapped in an AppWrapper being reset every 5 minutes (https://issues.redhat.com/browse/RHOAIENG-8834 and [discussed in #521](https://github.com/project-codeflare/codeflare-sdk/pull/521#pullrequestreview-2074255740)). 

The problem is that ODH fork of Kueue sets  `waitForPodsReady.enabled` to true (https://github.com/opendatahub-io/kueue/blob/dev/config/components/manager/controller_manager_config.yaml#L25-L26) while the AppWrapper controller was assuming that `waitForPodsReady.enabled` was false (which is the default in Kueue).   As a result, the AppWrapper Job Reconciler loop was not updating the `PodsReady` condition in the Workload object (this write is conditional on waitForPodsReady being true). 

We did a [quick fix](https://github.com/project-codeflare/appwrapper/pull/180) in AppWrapper 0.20.2.  The right longer term fix is probably for the codeflare operator to read the Kueue configuration on startup so we can ensure that the two parts of the system are in synch.


sutaakar (approved):
Tried locally, works well

PR #585, opened 2024-07-02
Run only GPU tests for PR check

sutaakar:
# Issue link

# What changes have been made

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/.github/workflows/e2e_tests.yaml
+++ b/.github/workflows/e2e_tests.yaml
@@ -56,6 +56,7 @@ jobs:
         uses: ./common/github-actions/nvidia-gpu-setup
 
       - name: Setup and start KinD cluster
+        id: kind-install
         uses: ./common/github-actions/kind
 
       - name: Install NVidia GPU operator for KinD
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e_tests.yaml
```diff
@@ -83,35 +84,34 @@ jobs:
           export CODEFLARE_TEST_TIMEOUT_GPU_PROVISIONING=30m
 
           export CODEFLARE_TEST_OUTPUT_DIR=${{ env.TEMP_DIR }}
-          echo "CODEFLARE_TEST_OUTPUT_DIR=${CODEFLARE_TEST_OUTPUT_DIR}" >> $GITHUB_ENV
 
           set -euo pipefail
-          go test -timeout 60m -v ./test/e2e -json 2>&1 | tee ${CODEFLARE_TEST_OUTPUT_DIR}/gotest.log | gotestfmt
+          go test -timeout 60m -v -skip "^Test.*Cpu$" ./test/e2e -json 2>&1 | tee ${CODEFLARE_TEST_OUTPUT_DIR}/gotest.log | gotestfmt
```
astefanutti, line 89: Curious to know if you prefer keeping two physical tests instead of one controlled with an environment variable. Is it for "reporting" purpose?

sutaakar: I don't have a strong opinion on that.
It is usable in downstream testing, where CPU and GPU tests are executed on different phases.
It also sufficiently distinguish those scenarios, so you can simply identify what option was executed from test name.

I kept it aligned with changes in https://github.com/opendatahub-io/distributed-workloads/pull/176

astefanutti: Sounds good, was more to get your opinion, thanks.

astefanutti (approved):

PR #589, opened 2024-07-10
make oauth image and cert generator image configurable via env vars

KPostOffice:
# Issue link
https://issues.redhat.com/browse/RHOAIENG-8552

# What changes have been made
I've made the cert generator image and oauthproxy image build time configurable

# Verification steps
run operator and verify that the images specified via make build are respected in the resulting RayCluster

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

# Note
There was also discussion about making these configurable at deployment time. I think the only difference would be whether we would need to re-release CFO to update the vs the RHODS/ODH operator. I lean towards leaving it on our end via build args but I'm open to suggestions

### DIFF
--- a/.github/workflows/project-codeflare-release.yml
+++ b/.github/workflows/project-codeflare-release.yml
@@ -98,7 +98,18 @@ jobs:
     steps:
     - name: Release CodeFlare operator
       run: |
-        gh workflow run tag-and-build.yml --repo ${{ github.event.inputs.codeflare-repository-organization }}/codeflare-operator --ref ${{ github.ref }} --field is-stable=${{ github.event.inputs.is-stable }} --field version=${{ github.event.inputs.operator-version }} --field replaces=${{ github.event.inputs.replaces }} --field codeflare-sdk-version=${{ github.event.inputs.codeflare-sdk-version }} --field appwrapper-version=${{ github.event.inputs.appwrapper-version }} --field kuberay-version=${{ github.event.inputs.kuberay-version }} --field kueue-version=${{ github.event.inputs.kueue-version }} --field quay-organization=${{ github.event.inputs.quay-organization }} --field community-operators-prod-fork-organization=${{ github.event.inputs.codeflare-repository-organization }} --field community-operators-prod-organization=${{ github.event.inputs.community-operators-prod-organization }}
+        gh workflow run tag-and-build.yml --repo ${{ github.event.inputs.codeflare-repository-organization }}/codeflare-operator \
+        --ref ${{ github.ref }} \
+        --field is-stable=${{ github.event.inputs.is-stable }} \
+        --field version=${{ github.event.inputs.operator-version }} \
+        --field replaces=${{ github.event.inputs.replaces }} \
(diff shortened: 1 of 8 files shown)


.github/workflows/project-codeflare-release.yml
```diff
@@ -22,6 +22,14 @@ on:
...
         description: 'Tested version of Kueue (for example: v0.0.0)'
         required: true
+      cert-generator-image:
+        description: 'Image used to generate RayCluster certificates'
+        required: true
+        default: 'registry.redhat.io/ubi9:latest'
+      oauth-proxy-image:
+        description: 'Image used to run OAuth proxy'
+        required: true
+        default: 'registry.redhat.io/openshift4/ose-oauth-proxy:latest'
```
sutaakar, line 32: If I remember correctly GitHub doesn't allow more than 10 input parameters for GH action.
Can we use defaults in the build?

Makefile
```diff
@@ -113,6 +113,9 @@ ifneq ($(shell git status --porcelain),)
 	BUILD_VERSION := $(BUILD_VERSION)-dirty
 endif
 
+OAUTH_IMAGE ?= registry.redhat.io/openshift4/ose-oauth-proxy:latest
+CERT_GENERATOR_IMAGE ?= registry.redhat.io/ubi9:latest
```
sutaakar, line 117: SHA is needed for disconnected clusters, so I think it should be preferred.

.github/workflows/project-codeflare-release.yml
```diff
@@ -98,7 +98,18 @@ jobs:
...
+        --field is-stable=${{ github.event.inputs.is-stable }} \
+        --field version=${{ github.event.inputs.operator-version }} \
+        --field replaces=${{ github.event.inputs.replaces }} \
+        --field codeflare-sdk-version=${{ github.event.inputs.codeflare-sdk-version }} \
+        --field appwrapper-version=${{ github.event.inputs.appwrapper-version }} \
+        --field kuberay-version=${{ github.event.inputs.kuberay-version }} \
+        --field kueue-version=${{ github.event.inputs.kueue-version }} \
+        --field quay-organization=${{ github.event.inputs.quay-organization }} \
+        --field community-operators-prod-fork-organization=${{ github.event.inputs.codeflare-repository-organization }} \
+        --field community-operators-prod-organization=${{ github.event.inputs.community-operators-prod-organization }}
```
sutaakar, line 112: AFAIK this change is not needed, right?

sutaakar:
Can you remove https://github.com/project-codeflare/codeflare-operator/blob/main/config/e2e/config.yaml#L10
And inject the image from ^ into env variable for e2e test, I suppose in https://github.com/project-codeflare/codeflare-operator/blob/main/config/e2e/patch_resources.yaml ?

.github/workflows/project-codeflare-release.yml
KPostOffice: It's not, but I left it in to improve readability. The line was extremely long

sutaakar (approved):

PR #592, opened 2024-07-18
Adjust e2e tests to verify deletion of resources

sutaakar:
# Issue link

# What changes have been made

Expand e2e tests to verify that resource deletion is successful.
Side reason for this change is to make sure e2e tests are stable as it seems that deleting Workload resources, local queue and namespace at the same time can lead to race condition when ClusterQueue is not properly updated, causing Workloads to hang.

# Verification steps

Verified by running e2e tests locally.

## Checks
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/test/e2e/mnist_rayjob_raycluster_test.go
+++ b/test/e2e/mnist_rayjob_raycluster_test.go
@@ -81,13 +81,6 @@ func runMnistRayJobRayCluster(t *testing.T, accelerator string, numberOfGpus int
 	test.T().Logf("Connecting to Ray cluster at: %s", rayDashboardURL.String())
 	rayClient := NewRayClusterClient(rayDashboardURL)
 
-	// Wait for Ray job id to be available, this value is needed for writing logs in defer
-	test.Eventually(RayJob(test, rayJob.Namespace, rayJob.Name), TestTimeoutShort).
-		Should(WithTransform(RayJobId, Not(BeEmpty())))
-
-	// Retrieving the job logs once it has completed or timed out
-	defer WriteRayJobAPILogs(test, rayClient, GetRayJobId(test, rayJob.Namespace, rayJob.Name))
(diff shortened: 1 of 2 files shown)


test/e2e/mnist_rayjob_raycluster_test.go
```diff
@@ -81,20 +81,22 @@ func runMnistRayJobRayCluster(t *testing.T, accelerator string, numberOfGpus int
...
 		To(WithTransform(RayJobStatus, Equal(rayv1.JobStatusSucceeded)))
+
+	WriteRayJobAPILogs(test, rayClient, GetRayJobId(test, rayJob.Namespace, rayJob.Name))
+
+	test.T().Logf("Deleting RayCluster %s/%s", rayCluster.Namespace, rayCluster.Name)
+	err = test.Client().Ray().RayV1().RayClusters(namespace.Name).Delete(test.Ctx(), rayCluster.Name, metav1.DeleteOptions{})
+	test.Expect(err).NotTo(HaveOccurred())
+
+	test.T().Logf("Waiting for RayCluster %s/%s to be deleted", rayCluster.Namespace, rayCluster.Name)
+	test.Eventually(RayClusters(test, namespace.Name), TestTimeoutLong).Should(BeEmpty())
```
jiripetrlik, line 99: Not sure if "TestTimeoutLong" is necessary here. Deleting resources is probably faster than creating new resources. Have you tried with shorter timeout?

sutaakar: hmm, I guess short timeout should be enough, will adjust the PR

sutaakar: Done.

jiripetrlik (approved):

PR #598, opened 2024-07-22
automatically inject default queue if not provided

KPostOffice:
# Issue link
https://issues.redhat.com/browse/RHOAIENG-6467

# What changes have been made
If no LocalQueue is specified by RayCluster and a default LocalQueue is specified, then the default LocalQueue will be added to the RayCluster spec

# Verification steps

1. Install Kueue, Ray, and this branch of the CFO
2. Create a default queue using the label {"kueue.x-k8s.io/default-queue": 'true'}
3. create a RayCluster with no LocalQueue specified
    1. see that the default queue is added to the RayCluster's labels
4. create a RayCluster with a LocalQueue specified
    1. see that the specified LocalQueue remains  

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/pkg/controllers/raycluster_webhook.go
+++ b/pkg/controllers/raycluster_webhook.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
@@ -45,6 +46,7 @@ var rayclusterlog = logf.Log.WithName("raycluster-resource")
 
 func SetupRayClusterWebhookWithManager(mgr ctrl.Manager, cfg *config.KubeRayConfiguration) error {
 	rayClusterWebhookInstance := &rayClusterWebhook{
+		Client: mgr.GetClient(),
 		Config: cfg,
 	}
 	return ctrl.NewWebhookManagedBy(mgr).
@@ -58,6 +60,7 @@ func SetupRayClusterWebhookWithManager(mgr ctrl.Manager, cfg *config.KubeRayConf
 // +kubebuilder:webhook:path=/validate-ray-io-v1-raycluster,mutating=false,failurePolicy=fail,sideEffects=None,groups=ray.io,resources=rayclusters,verbs=create;update,versions=v1,name=vraycluster.ray.openshift.ai,admissionReviewVersions=v1
 
 type rayClusterWebhook struct {
+	client.Client
 	Config *config.KubeRayConfiguration
 }
 
@@ -75,6 +78,12 @@ func (w *rayClusterWebhook) Default(ctx context.Context, obj runtime.Object) err
 		rayCluster.Spec.HeadGroupSpec.Template.Spec.Volumes = upsert(rayCluster.Spec.HeadGroupSpec.Template.Spec.Volumes, oauthProxyTLSSecretVolume(rayCluster), withVolumeName(oauthProxyVolumeName))
 
 		rayCluster.Spec.HeadGroupSpec.Template.Spec.ServiceAccountName = rayCluster.Name + "-oauth-proxy"
+	}
+
+	// add default queue label if not present
+	err := withDefaultLocalQueue(ctx, rayCluster, w.Client)
+	if err != nil {
+		return err
 	}
 
 	if ptr.Deref(w.Config.MTLSEnabled, true) {
--- a/pkg/controllers/raycluster_webhook_test.go
+++ b/pkg/controllers/raycluster_webhook_test.go
@@ -26,6 +26,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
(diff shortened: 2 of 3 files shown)


sutaakar:
@KPostOffice can you add unit tests in `raycluster_webhook_test.go`?

pkg/controllers/raycluster_webhook.go
```diff
@@ -77,6 +81,30 @@ func (w *rayClusterWebhook) Default(ctx context.Context, obj runtime.Object) err
 		rayCluster.Spec.HeadGroupSpec.Template.Spec.ServiceAccountName = rayCluster.Name + "-oauth-proxy"
 	}
 
+	// add default queue label if not present
```
dgrove-oss, line 84: If the RayCluster is owned by a Kueue-managed object (eg an AppWrapepr) then injecting the queue-name label on a RayCluster that does not have one is not correct.   This functionality at least needs to be guarded by configuration on the operator that allows it to be disabled.  To actually be robust it should be looking to see if the RayCluster is not owned by an AppWrapper.

dgrove-oss (changes requested):

pkg/controllers/raycluster_webhook.go
KPostOffice: I think this might actually warrant a webhook on the `Workload` CRD and checking if `spec.queue` is set. I don't understand why this approach is incorrect however, using a mutating webhook to inject defaults is a normal use case for mutating webhooks and shouldn't cause any issues since the value is changed prior to being applied to etcd meaning it shouldn't cause issues with re-reconciliation loops. Does this have something to do with AppWrapper functionality that I'm missing?

pkg/controllers/raycluster_webhook.go
```diff
@@ -77,6 +81,30 @@ func (w *rayClusterWebhook) Default(ctx context.Context, obj runtime.Object) err
...
+	if rayCluster.GetLabels() == nil {
+		rayCluster.Labels = make(map[string]string)
+	}
+	err := w.Client.List(ctx, &kueuev1beta1.LocalQueueList{})
+	if err != nil {
+		rayclusterlog.Error(err, "Failed to list LocalQueues, Kueue CRD might not be installed")
```
astefanutti, line 90: Kueue is optional, so the logic should pass-through the error.

pkg/controllers/raycluster_webhook.go
```diff
@@ -77,6 +81,30 @@ func (w *rayClusterWebhook) Default(ctx context.Context, obj runtime.Object) err
...
+		_, ok := rayCluster.Labels["kueue.x-k8s.io/queue-name"]
+		if !ok {
+			// check if CRD Kueue LocalQueue exists
+			localQueues := &kueuev1beta1.LocalQueueList{}
```
astefanutti, line 94: Could `metav1.PartialObjectMetadata` be used there, so it uses the partial metadata client / cache?

pkg/controllers/raycluster_webhook.go
```diff
@@ -77,6 +81,30 @@ func (w *rayClusterWebhook) Default(ctx context.Context, obj runtime.Object) err
...
+			err := w.Client.List(ctx, localQueues)
+			if err == nil {
+				for _, localQueue := range localQueues.Items {
+					is_default, ok := localQueue.Labels["kueue.x-k8s.io/default-queue"]
```
astefanutti, line 98: We should use something like `kueue.openshift.ai/default-queue` instead. We can also explore the idea upstream in parallel.

pkg/controllers/raycluster_webhook.go
dgrove-oss: My comment comes from Kueue's logic for supporting parent/child relationships for workloads.  It has code in its defaulting webhooks to detect that the object being created is a child of a Kueue managed parent object (encoded by a controller reference) and to not set the default value for `suspend` in a child job.  
https://github.com/kubernetes-sigs/kueue/blob/release-0.8/pkg/controller/jobframework/defaults.go#L23-L34

I'd suggest adding a similar check here to avoid adding a label which is going to have no impact at runtime and might be confusing to people.  If the ray cluster is controlled by an appwrapper, then it's the queue-name that is on the appwrapper that matters.  The queue-name (or lack of one) on the ray cluster does nothing. 

You can't easily call `IsOwnerManagedByKueue(owner)`, but you could either add a check for the appwrapper GVK or just check for whether the ray cluster is controller by anything at all (I think for standalone clusters, this should be no).

astefanutti: I think @dgrove-oss isn't questioning the use of the webhook, but raising the point the logic should only apply to the top-level resource managed by Kueue in the ownership graph.

astefanutti: I wonder if the functionality that this PR adds for RayCluster could be generalised and proposed upstream?

dgrove-oss: An upstream generalization would be nice.  For AppWrappers, we went with adding an optional defaultQueueName to the AppWrapper operator configuration.  It's less customizable (uses a single default queue name for all namespaces), but it covered the use case we had for MLBatch.  If an AppWrapper doesn't have a queue-name and the operator is configured with a default then the webhook injects it.  But it only works for AppWrappers.

KPostOffice: Ahh, I see thanks @dgrove-oss

pkg/controllers/raycluster_webhook.go
KPostOffice: This label is already being used in CF-sdk I'm not sure when it was added, but changing it could break users

dgrove-oss:
Should we close since local queue defaulting is now an alpha feature in upstream Kueue (0.10)?

KPostOffice:
> Should we close since local queue defaulting is now an alpha feature in upstream Kueue (0.10)?

Yeah, closing

PR #601, opened 2024-07-24
Add a check for head pod imagePullSecrets

Ygnas:
# Issue link

[https://issues.redhat.com/browse/RHOAIENG-9247](https://issues.redhat.com/browse/RHOAIENG-9247)
# What changes have been made

Added a check for imagePullSecrets in the head pod, and if it does not exists delete the head pod, so it can be redeployed.
# Verification steps

Set up an integrated OpenShift container registry and upload an image to it:
[https://github.com/opendatahub-io/distributed-workloads/tree/main/images/runtime/examples#readme](https://github.com/opendatahub-io/distributed-workloads/tree/main/images/runtime/examples#readme)
Use the uploaded image inside the notebooks, head pod should not enter the ImagePullBackOff state (happens only frequently but not always).

## Checks
- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/pkg/controllers/raycluster_controller.go
+++ b/pkg/controllers/raycluster_controller.go
@@ -213,6 +213,10 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 			return ctrl.Result{RequeueAfter: requeueTime}, err
 		}
 
+		if err := r.deleteHeadPodIfMissingImagePullSecrets(ctx, cluster); err != nil {
+			return ctrl.Result{RequeueAfter: requeueTime}, err
+		}
+
 		_, err = r.kubeClient.RbacV1().ClusterRoleBindings().Apply(ctx, desiredOAuthClusterRoleBinding(cluster), metav1.ApplyOptions{FieldManager: controllerName, Force: true})
 		if err != nil {
 			logger.Error(err, "Failed to update OAuth ClusterRoleBinding")
@@ -470,6 +474,7 @@ func generateCACertificate() ([]byte, []byte, error) {
 
 	return privateKeyPem, certPem, nil
 }
+
 func desiredWorkersNetworkPolicy(cluster *rayv1.RayCluster) *networkingv1ac.NetworkPolicyApplyConfiguration {
 	return networkingv1ac.NetworkPolicy(cluster.Name+"-workers", cluster.Namespace).
 		WithLabels(map[string]string{RayClusterNameLabel: cluster.Name}).
@@ -486,6 +491,7 @@ func desiredWorkersNetworkPolicy(cluster *rayv1.RayCluster) *networkingv1ac.Netw
 			metav1ac.OwnerReference().WithUID(cluster.UID).WithName(cluster.Name).WithKind(cluster.Kind).WithAPIVersion(cluster.APIVersion).WithController(true),
 		)
 }
+
 func desiredHeadNetworkPolicy(cluster *rayv1.RayCluster, cfg *config.KubeRayConfiguration, kubeRayNamespaces []string) *networkingv1ac.NetworkPolicyApplyConfiguration {
 	allSecuredPorts := []*networkingv1ac.NetworkPolicyPortApplyConfiguration{
 		networkingv1ac.NetworkPolicyPort().WithProtocol(corev1.ProtocolTCP).WithPort(intstr.FromInt(8443)),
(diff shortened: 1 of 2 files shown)


pkg/controllers/raycluster_controller.go
```diff
@@ -226,6 +226,10 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			return ctrl.Result{RequeueAfter: requeueTime}, err
 		}
 
+		if err := r.checkAndDeleteHeadPod(ctx, cluster); err != nil {
```
astefanutti, line 229: I'd suggest to move it close to the ServiceAccount apply call to make it clear that it's related.

pkg/controllers/raycluster_controller.go
```diff
@@ -544,6 +550,39 @@ func desiredHeadNetworkPolicy(cluster *rayv1.RayCluster, cfg *config.KubeRayConf
...
+			return fmt.Errorf("failed to delete head pod: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func getHeadPod(ctx context.Context, r *RayClusterReconciler, cluster *rayv1.RayCluster) (*corev1.Pod, error) {
+	podList, err := r.kubeClient.CoreV1().Pods(cluster.Namespace).List(ctx, metav1.ListOptions{
+		LabelSelector: "ray.io/node-type=head",
```
astefanutti, line 575: The cluster name label should also be added. There may be multiple Ray clusters in the namespace.

pkg/controllers/raycluster_controller.go
```diff
@@ -544,6 +550,39 @@ func desiredHeadNetworkPolicy(cluster *rayv1.RayCluster, cfg *config.KubeRayConf
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	if len(podList.Items) == 0 {
+		return nil, fmt.Errorf("head pod not found")
```
astefanutti, line 581: I'm not sure we should treat that case as an error?

pkg/controllers/raycluster_controller.go
```diff
@@ -544,6 +550,39 @@ func desiredHeadNetworkPolicy(cluster *rayv1.RayCluster, cfg *config.KubeRayConf
...
+	if err != nil {
+		return fmt.Errorf("failed to get OAuth ServiceAccount: %w", err)
+	}
+
+	headPod, err := getHeadPod(ctx, r, cluster)
+	if err != nil {
+		return fmt.Errorf("failed to get head pod: %w", err)
+	}
+
+	if len(headPod.Spec.ImagePullSecrets) < 1 && len(serviceAccount.ImagePullSecrets) > 0 {
```
astefanutti, line 564: It's probably more robust to check all the secrets from the ServiceAccount are present in the Pod.

sutaakar:
@Ygnas Can you provide unit tests for the change?

pkg/controllers/raycluster_controller.go
```diff
@@ -544,6 +550,51 @@ func desiredHeadNetworkPolicy(cluster *rayv1.RayCluster, cfg *config.KubeRayConf
...
+	}
+
+	headPod, err := getHeadPod(ctx, r, cluster)
+	if err != nil {
+		return fmt.Errorf("failed to get head pod: %w", err)
+	}
+
+	if serviceAccount.ImagePullSecrets == nil || len(serviceAccount.ImagePullSecrets) == 0 {
+		return fmt.Errorf("ServiceAccount missing ImagePullSecrets")
+	}
```
astefanutti, line 566: Is this really an error?

Ygnas: It is not an error, I though it needs to reconcile if serviceAccount has no ImagePullSecrets

pkg/controllers/raycluster_controller.go
```diff
@@ -544,6 +550,47 @@ func desiredHeadNetworkPolicy(cluster *rayv1.RayCluster, cfg *config.KubeRayConf
...
+	if err != nil {
+		return fmt.Errorf("failed to get OAuth ServiceAccount: %w", err)
+	}
+
+	headPod, err := getHeadPod(ctx, r, cluster)
+	if err != nil {
+		return fmt.Errorf("failed to get head pod: %w", err)
+	}
+
+	if headPod != nil {
```
astefanutti, line 564: I'd suggest to flip the logic, .e.g.:
```go
if headPod == nil {
  return nil
}
```

astefanutti:
/lgtm

I've tested it multiple times and never encountered the issue.

pkg/controllers/raycluster_controller.go
```diff
@@ -544,6 +550,49 @@ func desiredHeadNetworkPolicy(cluster *rayv1.RayCluster, cfg *config.KubeRayConf
 		)
 }
 
+func (r *RayClusterReconciler) checkAndDeleteHeadPod(ctx context.Context, cluster *rayv1.RayCluster) error {
```
astefanutti, line 553: nit: can it be renamed `deleteHeadPodIfMissingImagePullSecrets`?

pkg/controllers/raycluster_controller_test.go
```diff
@@ -145,6 +146,59 @@ var _ = Describe("RayCluster controller", func() {
...
+
+				if len(sa.ImagePullSecrets) == 0 {
+					sa.ImagePullSecrets = []corev1.LocalObjectReference{
+						{Name: "test-image-pull-secret"},
+					}
+					_, err = k8sClient.CoreV1().ServiceAccounts(namespaceName).Update(ctx, sa, metav1.UpdateOptions{})
+					if err != nil {
+						return err
+					}
+				}
```
sutaakar, line 189: This code doesn't belong into eventually check.
This code should be executed once SA is available.

Just a generic note - try to avoid conditions in tests if possible. In this case you can i.e. check that `ImagePullSecrets` is empty, as a next step you can put there a value.

Ygnas: Thanks I think, I made it way better now

pkg/controllers/raycluster_controller.go
```diff
@@ -213,6 +213,10 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			return ctrl.Result{RequeueAfter: requeueTime}, err
 		}
 
+		if err := r.deleteHeadPodIfMissingImagePullSecrets(ctx, cluster); err != nil {
+			return ctrl.Result{RequeueAfter: requeueTime}, err
+		}
```
KPostOffice, line 218: Can this not occur in WorkerPods too?

KPostOffice (commented):
Looks good to me, one point of clarification

pkg/controllers/raycluster_controller.go
astefanutti: Theoretically yes. Practically, the worker Pods init container that waits for the head Pod to become ready makes it so it does not happen. I'd be inclined to keep the logic simple at the moment and iterate if necessary.

pkg/controllers/raycluster_controller.go
```diff
@@ -544,6 +550,49 @@ func desiredHeadNetworkPolicy(cluster *rayv1.RayCluster, cfg *config.KubeRayConf
...
+	}
+	for _, secret := range headPod.Spec.ImagePullSecrets {
+		delete(missingSecrets, secret.Name)
+	}
+	if len(missingSecrets) > 0 {
+		if err := r.kubeClient.CoreV1().Pods(headPod.Namespace).Delete(ctx, headPod.Name, metav1.DeleteOptions{}); err != nil {
+			return fmt.Errorf("failed to delete head pod: %w", err)
+		}
+	}
+	return nil
```
zdtsw, line 580: I do not think i fully understand this part:
what would happen if the serviceAccount has multiple ImagePullSecrets? then it can get into an infinitely reconcile by deleting headPod?

Ygnas: @zdtsw I don't think it should, I just tried adding extra ImagePullSecrets into the SA, the head was deleted and recreated with the extra ones. I think all the secrets always end up in the headPod.

PR #603, opened 2024-07-30
Downgrade StatusReasonConflict errors to debug messages

tardieu:
The `codeflare-operator` log is littered with update conflict errors such as:
```2024-07-24T13:06:33Z    ERROR   Reconciler error        {"controller": "AppWrapper", "controllerGroup": "workload.codeflare.dev", "controllerKind": "AppWrapper", "AppWrapper": {"name":"kevin1-team-hw","namespace":"kevin1-team"}, "namespace": "kevin1-team", "name": "kevin1-team-hw", "reconcileID": "b6e57167-a357-4c67-85d1-f455e2b57ab6", "error": "Operation cannot be fulfilled on appwrappers.workload.codeflare.dev \"kevin1-team-hw\": the object has been modified; please apply your changes to the latest version and try again"}```

These update conflicts result from trying to update stale Kubernetes object revisions in etcd when multiple reconciliers (or users) are concurrently working on cached copies of these objects. These conflicts are harmless. They are handled by retrying the reconciliation loop, refreshing the cached object, and updating or patching the more recent revision. This process is entirely handled by the controller runtime but it involves returning the conflict error to the controller runtime to trigger these retries. Unfortunately, the controller runtime as a result unconditionally logs these harmless conflicts as errors, which is confusing users.

This PR therefore wraps the controller runtime logger with a filter that downgrades these log messages from `ERROR` to `DEBUG` messages, more accurately matching the gravity of the event.

### DIFF
--- a/main.go
+++ b/main.go
@@ -116,7 +116,7 @@ func main() {
 	zapOptions.BindFlags(flag.CommandLine)
 	flag.Parse()
 
-	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&zapOptions)))
+	ctrl.SetLogger(controllers.FilteredLogger(zap.New(zap.UseFlagOptions(&zapOptions))))
 	klog.SetLogger(ctrl.Log)
 
 	setupLog.Info("Build info",
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/project-codeflare/codeflare-operator
 go 1.22.2
 
 require (
+	github.com/go-logr/logr v1.4.2
 	github.com/onsi/ginkgo/v2 v2.19.0
 	github.com/onsi/gomega v1.33.1
 	github.com/open-policy-agent/cert-controller v0.10.1
@@ -51,7 +52,6 @@ require (
 	github.com/evanphx/json-patch v5.6.0+incompatible // indirect
 	github.com/evanphx/json-patch/v5 v5.8.0 // indirect
 	github.com/fsnotify/fsnotify v1.7.0 // indirect
-	github.com/go-logr/logr v1.4.2 // indirect
 	github.com/go-logr/zapr v1.3.0 // indirect
 	github.com/go-openapi/jsonpointer v0.20.0 // indirect
 	github.com/go-openapi/jsonreference v0.20.2 // indirect
--- a/pkg/controllers/support.go
+++ b/pkg/controllers/support.go
@@ -3,11 +3,13 @@ package controllers
 import (
 	"os"
 
+	"github.com/go-logr/logr"
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
 
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: main.go [hunk 1]
```diff
@@ -116,7 +116,7 @@ func main() {
 	zapOptions.BindFlags(flag.CommandLine)
 	flag.Parse()
 
-	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&zapOptions)))
+	ctrl.SetLogger(controllers.FilteredLogger(zap.New(zap.UseFlagOptions(&zapOptions))))
```
varshaprasad96, line 119: Given these are reconciler errors, being returned from the server after a client operation - would it be possible to catch them after every patch/update operations, using `apierrors.IsConflict` (https://github.com/kubernetes/apimachinery/blob/95b78024e3feada7739b40426690b4f287933fd8/pkg/api/errors/errors.go#L546) rather than creating a new logger for it. 

```go
if apierrors.IsConflict(err) {
            log.V(1).Info("Conflict error during reconciliation", "error", err)
            return nil
 } 
return err
```

This way the logging level can also be modified, for us to be able to avoid them if required. 

We should be able to catch this for client side operations iiuc, as we do in controller-runtime wherein there are tests to ensure that this appears during a version mismatch:
https://github.com/kubernetes-sigs/controller-runtime/blob/67854429a3437a7684136c46fd8af1d33ea90de4/pkg/client/fake/client_test.go#L680

varshaprasad96: Also, just out of curiosity, though these conflict errors are benign, its generally a good practice to do a `get` followed by `update` in a reconciler (https://github.com/operator-framework/operator-sdk/issues/3968#issuecomment-705688365). Are we expecting these errors by design, or a better solution around them is to actually use SSA?

tardieu: The AppWrapper reconciler follows the `get` -> `update` pattern. However this `get` fetches the object from the controller cache, which might be stale. Even if we were to pull from etcd directly (which is strongly discouraged due to the added etcd load) this does not guarantee freedom from concurrent updates initiated by other reconcilers or users.

tardieu: In the long run, I agree we would like to address the issue at the source by not returning Conflict errors in the first place. However, this requires changes to the upstream Kueue project (from which the GenericReconcilier is instantiated). Until such changes are approved, implemented, and released, this PR is the least intrusive way to fix the logging problem and address the concerns of our cluster admins.

varshaprasad96: @tardieu Thanks for explaining. Just to understand this better, could you elaborate on why the change needs to be done in the Kueue project. The log here indicates that this is from the AppWrapper controller. Wouldn't modifying just the AppWrapper controller be enough?

varshaprasad96: > However, the problematic reconciler I am referring to is instantiated in the appwrapper code (https://github.com/project-codeflare/appwrapper/blob/main/internal/controller/workload/workload_controller.go#L49) but implemented in Kueue (https://github.com/kubernetes-sigs/kueue/blob/v0.8.0/pkg/controller/jobframework/reconciler.go#L1086).

Ah okay! This was helpful in clearing it out! Thanks @tardieu!

varshaprasad96 (commented):
lgtm

PR #605, opened 2024-08-06
add hash to end of resource names to avoid name clash

KPostOffice:
# Issue link
https://issues.redhat.com/browse/RHOAIENG-10129

# What changes have been made
created a function which takes a string and appends the controller name to it then hashes it to use as a suffix for the resources created by this controller to avoid naming conflicts

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/pkg/controllers/raycluster_controller.go
+++ b/pkg/controllers/raycluster_controller.go
@@ -114,6 +114,12 @@ var (
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.15.3/pkg/reconcile
 
+func shouldUseOldName(cluster *rayv1.RayCluster) bool {
+	// hashed name code was added in the same commit as the version annotation
+	_, ok := cluster.GetAnnotations()[versionAnnotation]
+	return !ok
+}
+
 func (r *RayClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	logger := ctrl.LoggerFrom(ctx)
 
@@ -304,7 +310,10 @@ func isMTLSEnabled(cfg *config.KubeRayConfiguration) bool {
 }
 
 func crbNameFromCluster(cluster *rayv1.RayCluster) string {
-	return cluster.Name + "-" + cluster.Namespace + "-auth" // NOTE: potential naming conflicts ie {name: foo, ns: bar-baz} and {name: foo-bar, ns: baz}
+	if shouldUseOldName(cluster) {
+		return cluster.Name + "-" + cluster.Namespace + "-auth"
+	}
+	return RCCUniqueName(cluster.Name + "-" + cluster.Namespace + "-auth")
 }
 
 func desiredOAuthClusterRoleBinding(cluster *rayv1.RayCluster) *rbacv1ac.ClusterRoleBindingApplyConfiguration {
@@ -326,7 +335,10 @@ func desiredOAuthClusterRoleBinding(cluster *rayv1.RayCluster) *rbacv1ac.Cluster
 }
 
 func oauthServiceAccountNameFromCluster(cluster *rayv1.RayCluster) string {
-	return cluster.Name + "-oauth-proxy"
+	if shouldUseOldName(cluster) {
+		return cluster.Name + "-oauth-proxy"
+	}
+	return RCCUniqueName(cluster.Name + "-oauth-proxy")
 }
 
 func desiredServiceAccount(cluster *rayv1.RayCluster) *corev1ac.ServiceAccountApplyConfiguration {
@@ -363,11 +375,17 @@ func desiredClusterRoute(cluster *rayv1.RayCluster) *routev1ac.RouteApplyConfigu
 }
 
 func oauthServiceNameFromCluster(cluster *rayv1.RayCluster) string {
-	return cluster.Name + "-oauth"
+	if shouldUseOldName(cluster) {
+		return cluster.Name + "-oauth"
+	}
+	return RCCUniqueName(cluster.Name + "-oauth")
 }
 
 func oauthServiceTLSSecretName(cluster *rayv1.RayCluster) string {
-	return cluster.Name + "-proxy-tls-secret"
+	if shouldUseOldName(cluster) {
+		return cluster.Name + "-proxy-tls-secret"
+	}
+	return RCCUniqueName(cluster.Name + "-proxy-tls-secret")
 }
 
 func desiredOAuthService(cluster *rayv1.RayCluster) *corev1ac.ServiceApplyConfiguration {
@@ -389,7 +407,10 @@ func desiredOAuthService(cluster *rayv1.RayCluster) *corev1ac.ServiceApplyConfig
 }
 
 func oauthSecretNameFromCluster(cluster *rayv1.RayCluster) string {
-	return cluster.Name + "-oauth-config"
+	if shouldUseOldName(cluster) {
+		return cluster.Name + "-oauth-config"
+	}
+	return RCCUniqueName(cluster.Name + "-oauth-config")
 }
 
 // desiredOAuthSecret defines the desired OAuth secret object
@@ -406,7 +427,10 @@ func desiredOAuthSecret(cluster *rayv1.RayCluster, cookieSalt string) *corev1ac.
 }
 
 func caSecretNameFromCluster(cluster *rayv1.RayCluster) string {
-	return "ca-secret-" + cluster.Name
+	if shouldUseOldName(cluster) {
+		return "ca-secret-" + cluster.Name
+	}
+	return RCCUniqueName(cluster.Name + "-ca-secret")
 }
 
 func desiredCASecret(cluster *rayv1.RayCluster, key, cert []byte) *corev1ac.SecretApplyConfiguration {
@@ -462,8 +486,17 @@ func generateCACertificate() ([]byte, []byte, error) {
 	return privateKeyPem, certPem, nil
 }
 
+func workerNWPNameFromCluster(cluster *rayv1.RayCluster) string {
+	if shouldUseOldName(cluster) {
+		return cluster.Name + "-workers"
+	}
+	return RCCUniqueName(cluster.Name + "-workers")
+}
+
 func desiredWorkersNetworkPolicy(cluster *rayv1.RayCluster) *networkingv1ac.NetworkPolicyApplyConfiguration {
-	return networkingv1ac.NetworkPolicy(cluster.Name+"-workers", cluster.Namespace).
+	return networkingv1ac.NetworkPolicy(
+		workerNWPNameFromCluster(cluster), cluster.Namespace,
+	).
 		WithLabels(map[string]string{RayClusterNameLabel: cluster.Name}).
 		WithSpec(networkingv1ac.NetworkPolicySpec().
 			WithPodSelector(metav1ac.LabelSelector().WithMatchLabels(map[string]string{"ray.io/cluster": cluster.Name, "ray.io/node-type": "worker"})).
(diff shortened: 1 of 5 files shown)


pkg/controllers/raycluster_controller.go
```diff
@@ -343,11 +344,11 @@ func desiredServiceAccount(cluster *rayv1.RayCluster) *corev1ac.ServiceAccountAp
 }
 
 func dashboardNameFromCluster(cluster *rayv1.RayCluster) string {
-	return "ray-dashboard-" + cluster.Name
+	return "ray-dashboard-" + cluster.Name + "-" + ControllerSpecificHash(cluster.Name, defaultHashLength)
 }
 
 func rayClientNameFromCluster(cluster *rayv1.RayCluster) string {
-	return "rayclient-" + cluster.Name
+	return "rayclient-" + cluster.Name + "-" + ControllerSpecificHash(cluster.Name, defaultHashLength)
```
astefanutti, line 351: To go a step further, what about having a function is that file that returns the unique name, provided the resource name?

pkg/controllers/support.go
```diff
@@ -212,3 +214,15 @@ func (l logSink) WithName(name string) logr.LogSink {
 func FilteredLogger(logger logr.Logger) logr.Logger {
 	return logger.WithSink(logSink{logger.GetSink()})
 }
+
+var (
+	hashConstant = "codeflare-operator"
+)
+
+func ControllerSpecificHash(s string, l int) string {
```
astefanutti, line 222: I'd be tempted to not have the hash length as an argument. WDYT?

astefanutti (commented):
Ideally the existing resources should be pruned, like during the operator upgrade. But I think that's acceptable.

pkg/controllers/support.go
KPostOffice: In that case should we just leave it defined by a variable at `8`?

KPostOffice:
> Ideally the existing resources should be pruned, like during the operator upgrade. But I think that's acceptable.

I had this thought too. I can try my hand at writing an upgrade strategy for this

pkg/controllers/raycluster_controller.go
KPostOffice: so do something that basically does:
```
getUniqueName(name string) string {
  return name + "-" + ControllerSpecificHash(name)
}
```

astefanutti: Yes, exactly.

pkg/controllers/support.go
astefanutti: Yes, like adding a `hashLength` variable along side the `hashConstant`.

pkg/controllers/raycluster_controller.go
```diff
@@ -406,7 +410,7 @@ func desiredOAuthSecret(cluster *rayv1.RayCluster, cookieSalt string) *corev1ac.
 }
 
 func caSecretNameFromCluster(cluster *rayv1.RayCluster) string {
-	return "ca-secret-" + cluster.Name
+	return getUniqueName("ca-secret-" + cluster.Name)
```
astefanutti, line 413: Maybe seize that opportunity to be consistent with the other resources, i.e. `cluster.name` + "-ca-secret`?

pkg/controllers/raycluster_controller.go
```diff
@@ -619,3 +625,45 @@ func (r *RayClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 	return controller.Complete(r)
 }
+
+func deleteDeprecatedObjects(ctx context.Context, r *RayClusterReconciler, cluster *rayv1.RayCluster) error {
+	// Delete deprecated objects if they exist. These have all been replace by objects with names generated by
```
astefanutti, line 630: Maybe add a TODO to have that logic removed after a reasonable number of releases?

pkg/controllers/raycluster_controller.go
KPostOffice: Consider it seized

pkg/controllers/support.go
```diff
@@ -209,3 +211,16 @@ func ownerRefForRayCluster(cluster *rayv1.RayCluster) *v1.OwnerReferenceApplyCon
 		WithUID(cluster.UID).
 		WithController(true)
 }
+
+var (
+	hashConstant = "codeflare-operator"
```
astefanutti, line 216: I'd suggest `ray-cluster-controller` to be more portable.

astefanutti:
I guess the new names should be reflected into Ray cluster the webhook.

astefanutti:
Overall LGTM, just left a comment on the hash seed, and to update the Ray cluster webhook accordingly.

pkg/controllers/support.go
KPostOffice: I did this based on it being non-specific to a controller. It could also be used for the appwrapper controller. I can make it something like this instead:

```
func seededHash(seed string, s string) string {
    return fmt.Sprintf("%x", sha256.Sum256([]byte(seed+s)))[:hashLength]
}
```

and then create a function in the rayClusterController:

```
func rccUniqueName(s string) string {
    return s + "-" + seededHash("ray-cluster-controller", s)
```

KPostOffice:
> I guess the new names should be reflected into Ray cluster the webhook.

I think this is taken care of already by using the `nameFromCluster` functions in all the necessary places.

pkg/controllers/support.go
astefanutti: Ah, actually my thinking is to make it specific to the Ray cluster controller so names won't change if we move that controller outside the codeflare operator. Having said that, your suggestion is a good alternative.

KPostOffice:
> > > I guess the new names should be reflected into Ray cluster the webhook.
> > 
> > 
> > I think this is taken care of already by using the `nameFromCluster` functions in all the necessary places.
> 
> Ah OK, I thought it could explain why the e2e failed.

The tests also needed to be changed to use the `fromCluster` functions. I'm still messing around with them. I thought I caught everything but I'm still clearly missing something that needs to be fixed in the tests

KPostOffice:
@astefanutti This should be g2g now

pkg/controllers/raycluster_controller.go
```diff
@@ -619,3 +626,47 @@ func (r *RayClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 	return controller.Complete(r)
 }
+
+func RCCUniqueName(s string) string {
+	return s + "-" + seededHash(controllerName, s)
+}
+
+func deleteDeprecatedObjects(ctx context.Context, r *RayClusterReconciler, cluster *rayv1.RayCluster) error {
```
astefanutti, line 634: For existing Ray clusters, is my understanding correct the existing resources will be deleted once the operator is upgraded, but these existing Ray clusters will still point to those deleted resources?

KPostOffice: Ah, yeah. I'll go this route. It should be pretty straight forward

pkg/controllers/raycluster_webhook.go
```diff
@@ -68,13 +69,21 @@ var _ webhook.CustomValidator = &rayClusterWebhook{}
 func (w *rayClusterWebhook) Default(ctx context.Context, obj runtime.Object) error {
 	rayCluster := obj.(*rayv1.RayCluster)
 
+	// add annotation to use new names
+	annotations := rayCluster.GetAnnotations()
+	if annotations == nil {
+		annotations = make(map[string]string)
+	}
+	annotations[newNameAnnotation] = "true"
```
astefanutti, line 77: While we are at it, it could be useful to generalise that annotation so that it contains the version of the operator that defaulted the Ray cluster. WDYT?

pkg/controllers/raycluster_webhook.go
```diff
@@ -58,23 +60,33 @@ func SetupRayClusterWebhookWithManager(mgr ctrl.Manager, cfg *config.KubeRayConf
...
 func (w *rayClusterWebhook) Default(ctx context.Context, obj runtime.Object) error {
+	logger := ctrl.LoggerFrom(ctx)
 	rayCluster := obj.(*rayv1.RayCluster)
 
+	// add annotation to use new names
+	annotations := rayCluster.GetAnnotations()
+	if annotations == nil {
+		annotations = make(map[string]string)
+	}
+	annotations[versionAnnotation] = "w.OperatorVersion"
```
sutaakar, line 80:
```suggestion
	annotations[versionAnnotation] = w.OperatorVersion
```

pkg/controllers/raycluster_webhook_test.go
```diff
@@ -46,6 +46,9 @@ func TestRayClusterWebhookDefault(t *testing.T) {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      rayClusterName,
 			Namespace: namespace,
+			Annotations: map[string]string{
+				versionAnnotation: "0.0.0",
+			},
```
sutaakar, line 51: Can you remove this annotation from here?
RayCluster CR created by used or SDK won't have the annotation defined. TO reproduce the behavior we shouldn't provide it here either.

pkg/controllers/raycluster_webhook_test.go
```diff
@@ -46,6 +46,9 @@ func TestRayClusterWebhookDefault(t *testing.T) {
 		ObjectMeta: metav1.ObjectMeta{
```
sutaakar, line 46: Can you set operator version into `rcWebhook` above?

pkg/controllers/raycluster_webhook_test.go
```diff
@@ -124,7 +127,7 @@ func TestRayClusterWebhookDefault(t *testing.T) {
 
 	t.Run("Expected required service account name for the head group", func(t *testing.T) {
 		test.Expect(validRayCluster.Spec.HeadGroupSpec.Template.Spec.ServiceAccountName).
-			To(Equal(validRayCluster.Name+"-oauth-proxy"),
+			To(Equal(oauthServiceAccountNameFromCluster(validRayCluster)),
```
sutaakar, line 130: Personally I would prefer using specific string here instead of function.
Using function can potentially hide expected value, as you can't be sure if the test passed because expected value is `validRayCluster.Name+"-oauth-proxy"` or `validRayCluster.Name+"-oauth-proxy-ecaffa62"`

sutaakar: And below too.

pkg/controllers/raycluster_webhook_test.go
```diff
@@ -230,11 +233,23 @@ func TestRayClusterWebhookDefault(t *testing.T) {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      rayClusterName,
+			Namespace: namespace,
+			Annotations: map[string]string{
+				versionAnnotation: "0.0.0",
+			},
+		},
+		Spec: rayv1.RayClusterSpec{},
+	}
 	validRayCluster := &rayv1.RayCluster{
```
sutaakar, line 246: _no action comment_
Using annotation and functions for `validRayCluster` is correct as those values are set by default webhook.

pkg/controllers/raycluster_webhook_test.go
```diff
@@ -351,11 +366,25 @@ func TestValidateCreate(t *testing.T) {
 
 func TestValidateUpdate(t *testing.T) {
```
sutaakar, line 367: It could be good to keep test case to validate update cluster without annotation and add annotation one as new test case.

pkg/controllers/raycluster_controller.go
```diff
@@ -304,7 +310,10 @@ func isMTLSEnabled(cfg *config.KubeRayConfiguration) bool {
 }
 
 func crbNameFromCluster(cluster *rayv1.RayCluster) string {
-	return cluster.Name + "-" + cluster.Namespace + "-auth" // NOTE: potential naming conflicts ie {name: foo, ns: bar-baz} and {name: foo-bar, ns: baz}
+	if shouldUseHashedName(cluster) {
```
astefanutti, line 313: I'd suggest to flip the check, so the main / new logic branch is outside the if block.

sutaakar (commented):
/lgtm

astefanutti (approved):

PR #609, opened 2024-08-20
codflare-operator manifests kustomize v5 upgrade

oksanabaza:
# Issue link

# What changes have been made

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


config/manifests/config-manifests-kustomization-control-test.yaml
```diff
@@ -0,0 +1,1219 @@
+apiVersion: v1
```
astefanutti, line 1: Is this file necessary?

config/manifests/config-manifests-kustomization-test.yaml
```diff
@@ -0,0 +1,1219 @@
+apiVersion: v1
```
astefanutti, line 1: Is this file necessary?

config/manifests/config-manifests-kustomization-control-test.yaml
oksanabaza: we don't need this file as this is an output of kustomize, will delete it, thanks

config/manifests/config-manifests-kustomization-test.yaml
oksanabaza: same here, thanks

astefanutti:
@oksanabaza there is one pre-commit check that fails (EOF), once it's fixed it should be good to go.

Bobbins228 (approved):
/lgtm 
Tested and works as expected nice work Oksana.

astefanutti (approved):

PR #613, opened 2024-09-03
RBAC fix to enable slack cluster queue lending limit adjustment

dgrove-oss:
The codeflare operator needs permission to read and write clusterqueues to enable the AppWrapper controller to adjust the lending limit of a designated slack cluster queue to reflect cordoned nodes.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -171,6 +171,16 @@ rules:
 - apiGroups:
   - kueue.x-k8s.io
   resources:
+  - clusterqueues
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - kueue.x-k8s.io
+  resources:
   - resourceflavors
   verbs:
   - get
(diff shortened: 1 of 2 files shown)


dgrove-oss:
Although we can work around this in MLBatch, it would be nice if this fix could be merged in time to make the next release so MLBatch only needs to have the patch for the codeflare operator's role in our configuration for RHOAI 2.12.

varshaprasad96 (commented):
@dgrove-oss The changes seem good to provide right permissions to the slack CQ. 

I assume this is only for the purposes of managing the Slack queue (as mentioned in the description). However, do we have admission policy (or a validating webhook) in place to not allow modifications to the slack CQ by anyone other than the app wrapper controller (would need a SA for this). Just concerned that we are not letting the same CQ be modified by Kueue as well as appwrapper controllers.

dgrove-oss:
Summarizing an offline discussion, the `lendingLimit` field of the ClusterQueue won't be updated by the kueue controller.  Cluster Admins in MLBatch are expected to modify the quota information in the slack ClusterQueue, but are expecting the AppWrapper controller to be modifying the lendingLimit.   The AppWrapper controller re-computes the value of the lendingLimit each time from the quota and the node status, so even if the value was mistakenly modified by a cluster admin the controller would correct the mistake on the next reconcile.

varshaprasad96 (approved):
/lgtm
/approve

PR #616, opened 2024-09-26
Add creation of kueue resources in e2e tests

ChughShilpa:
# Issue link
[RHOAIENG-11047](https://issues.redhat.com/browse/RHOAIENG-11047)

# What changes have been made
Since we are adjusting e2e tests to run with AMD gpu, thus 
- Adding creation of kueue resources as part of test to make it compatible with gpuresources (which is covered in next PR)
- Removing creation of kueue resources  from setup script
- Adjust methods with updated Ray cluster client

# Verification steps
Run test manually 
`go test -timeout 30m -v ./test/e2e -run TestMnistPyTorchAppWrapperCpu`

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [X] Manual tests

### DIFF
--- a/test/e2e/mnist_pytorch_appwrapper_test.go
+++ b/test/e2e/mnist_pytorch_appwrapper_test.go
@@ -28,23 +28,30 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/kueue/apis/kueue/v1beta1"
 )
 
 func TestMnistPyTorchAppWrapperCpu(t *testing.T) {
-	runMnistPyTorchAppWrapper(t, "cpu")
+	runMnistPyTorchAppWrapper(t, "cpu", 0)
 }
 
 func TestMnistPyTorchAppWrapperGpu(t *testing.T) {
-	runMnistPyTorchAppWrapper(t, "gpu")
+	runMnistPyTorchAppWrapper(t, "gpu", 1)
 }
 
 // Trains the MNIST dataset as a batch Job in an AppWrapper, and asserts successful completion of the training job.
-func runMnistPyTorchAppWrapper(t *testing.T, accelerator string) {
+func runMnistPyTorchAppWrapper(t *testing.T, accelerator string, numberOfGpus int) {
 	test := With(t)
 
-	// Create a namespace and localqueue in that namespace
+	// Create a namespace
 	namespace := test.NewTestNamespace()
-	localQueue := CreateKueueLocalQueue(test, namespace.Name, "e2e-cluster-queue")
+
+	// Create Kueue resources
+	resourceFlavor := CreateKueueResourceFlavor(test, v1beta1.ResourceFlavorSpec{})
+	defer test.Client().Kueue().KueueV1beta1().ResourceFlavors().Delete(test.Ctx(), resourceFlavor.Name, metav1.DeleteOptions{})
+	clusterQueue := createClusterQueue(test, resourceFlavor, numberOfGpus)
+	defer test.Client().Kueue().KueueV1beta1().ClusterQueues().Delete(test.Ctx(), clusterQueue.Name, metav1.DeleteOptions{})
+	localQueue := CreateKueueLocalQueue(test, namespace.Name, clusterQueue.Name, AsDefaultQueue)
 
 	// Test configuration
 	config := &corev1.ConfigMap{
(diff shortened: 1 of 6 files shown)


test/e2e/mnist_rayjob_raycluster_test.go
```diff
@@ -408,8 +420,42 @@ func getRayDashboardURL(test Test, namespace, rayClusterName string) url.URL {
...
+						Resources: []v1beta1.ResourceQuota{
+							{
+								Name:         corev1.ResourceCPU,
+								NominalQuota: resource.MustParse("8"),
+							},
+							{
+								Name:         corev1.ResourceMemory,
+								NominalQuota: resource.MustParse("12Gi"),
+							},
+							{
```
astefanutti, line 449: Should it be removed if `numberOfGpus` is 0?

test/e2e/mnist_rayjob_raycluster_test.go
```diff
@@ -408,8 +420,42 @@ func getRayDashboardURL(test Test, namespace, rayClusterName string) url.URL {
...
+								Name:         corev1.ResourceName("nvidia.com/gpu"),
+								NominalQuota: resource.MustParse(fmt.Sprint(numberOfGpus)),
+							},
+						},
+					},
+				},
+			},
+		},
 	}
+	clusterQueue := CreateKueueClusterQueue(test, cqSpec)
```
astefanutti, line 459:
```suggestion
	return CreateKueueClusterQueue(test, cqSpec)
```

test/e2e/mnist_pytorch_appwrapper_test.go
```diff
@@ -31,20 +32,26 @@ import (
...
 }
 
 func TestMnistPyTorchAppWrapperGpu(t *testing.T) {
-	runMnistPyTorchAppWrapper(t, "gpu")
+	runMnistPyTorchAppWrapper(t, "gpu", 1)
 }
 
 // Trains the MNIST dataset as a batch Job in an AppWrapper, and asserts successful completion of the training job.
-func runMnistPyTorchAppWrapper(t *testing.T, accelerator string) {
+func runMnistPyTorchAppWrapper(t *testing.T, accelerator string, numberOfGpus int) {
```
astefanutti, line 43: It's be good we move to an `Option` variadic argument.

test/e2e/mnist_rayjob_raycluster_test.go
ChughShilpa: IIRC specifying the resource name is mandatory even if nominalquota is 0

ChughShilpa: https://docs.redhat.com/en/documentation/red_hat_openshift_ai_self-managed/2-latest/html/working_with_distributed_workloads/configuring-distributed-workloads_distributed-workloads#configuring-quota-management-for-distributed-workloads_distributed-workloads:~:text=Specify%20the%20resource%20name%20and%20nominalQuota%20in%20the%20flavors.resources%20section%2C%20even%20if%20the%20nominalQuota%20value%20is%200.

sutaakar: The cleanest solution could be to move creation of ResourceFlavor and ClusterQueue (in the end also RayCluster) into `Test<something>Gpu/Cpu` functions. Each test function would then set properties relevant just to the test function.
That way we would get rid of conditions in test, reducing its reading complexity. Though that would mean a large rewrite.
I am ok with current solution too, just thinking out loud.

sutaakar: > IIRC specifying the resource name is mandatory even if nominalquota is 0

I think it is (was?) mandatory in RHOAI due to SDK passing GPU values. Not sure if it is still needed.
Anyway in real use cases for finetuning users need GPU.

ChughShilpa: Created a jira issue for SDK to address this - https://issues.redhat.com/browse/RHOAIENG-13618 
@astefanutti @sutaakar

test/e2e/mnist_pytorch_appwrapper_test.go
ChughShilpa: Thanks for the suggestion @astefanutti 
Will apply this suggestion in a different PR.

sutaakar (commented):
/lgtm

ChughShilpa:
@sutaakar @astefanutti 
Could someone please approve the PR ?

sutaakar:
> We may want to think about a strategy for the creation of cluster resources like ClusterQueues in the context of concurrent execution of tests in the future.

Current strategy is to provide generated name for cluster resources.

PR #617, opened 2024-09-26
Update appwrappers to v0.25.0

dgrove-oss:
Changes:
+ Add age printcolumn
+ Update to Kueue 0.8.1
+ autopilot: add TESTING NoSchedule taint to default config


go.mod
```diff
@@ -35,10 +35,13 @@ replace go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp => go.open
...
-// These replace directives support the backlevel go version required by ODH build
-replace github.com/project-codeflare/appwrapper v0.24.0 => github.com/project-codeflare/appwrapper v0.24.1-0.20240828152845-bfaac68eef30
+// This replace directive supports the backlevel go version required by the ODH build
+replace github.com/project-codeflare/appwrapper v0.25.0 => github.com/project-codeflare/appwrapper v0.25.1-0.20240926155059-30a8af17b8f4
 
-replace sigs.k8s.io/kueue v0.8.0 => github.com/opendatahub-io/kueue v0.7.0-odh-test
+// These replace directives deal with the backlevel ODH kueue version
+replace sigs.k8s.io/kueue v0.8.1 => github.com/opendatahub-io/kueue v0.7.0-odh-2
+
+replace github.com/prometheus/common v0.57.0 => github.com/prometheus/common v0.46.0
```
sutaakar, line 44: Is this needed just because `codeflare-common` depends on older version of prometheus or are there other components requiring this?

Fiona-Waters (commented):
/lgtm

sutaakar (approved):

go.mod
dgrove-oss: The prometheus/common replace is because of codeflare-common.  It probably should be updated there but I'm not sure how far that change would ripple.

sutaakar: ok, I think the change scope should be limited
will update it

sutaakar: PR ready - https://github.com/project-codeflare/codeflare-common/pull/70

dgrove-oss: PR #619 to finish the cleanup.

PR #621, opened 2024-10-16
Update appwrappers to v0.26.0

dgrove-oss:
Changes:
  + Redesign node monitoring to account for Node deletion
  + Also log the delta when lendingLimit is updated
  + Implement spec.managedBy field


Fiona-Waters (commented):
/lgtm

PR #623, opened 2024-10-31
Add RBAC aggregator cluster roles for appwrappers

dgrove-oss:


dgrove-oss:
@Bobbins228

Please validate that added clusterroles get properly propagated through the downstream builds to RHOAI.  I don't have visibility into that process.

Bobbins228:
@dgrove-oss On next release the changes will be automatically propagated downstream but I will keep an eye on that just in case :+1:

sutaakar:
/lgtm
Used https://github.com/project-codeflare/codeflare-operator/pull/624 to check that namespaced admin can work with AppWrapper by default.

PR #628, opened 2024-11-15
Raise Kueue version in CodeFlare operator

sutaakar:
# Issue link

N/A

# What changes have been made

Raise Kueue version in CodeFlare operator to use latest Kueue dependency available in ODH.

# Verification steps

E2E tests pass.

## Checks
- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [ ] Testing is not required for this change

PR #629, opened 2024-11-18
Update Kueue dependency to 0.8.3

dgrove-oss:
Update ODH fork of Kueue to use v0.8.3 tag
Update kuberay version to match that used in Kueue v0.8.3.


dgrove-oss:
Closing in favor of #628;  this was an attempt to help debug the test case failure that is being fixed in #630

PR #630, opened 2024-11-19
AppWrapper e2e: Remove creationTimestamp from marshalled template

sutaakar:
# Issue link

# What changes have been made

Remove creationTimestamp from marshalled template for AppWrapper. This creationTimestamp is brought by marshaller (it is not omitted because it is defined as struct instead of pointer in k8s API). This field breaks Kueue reconciliation in AppWrapper for Kueue 0.8.3.

# Verification steps

The change was verified in https://github.com/project-codeflare/codeflare-operator/pull/628

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/test/e2e/mnist_pytorch_appwrapper_test.go
+++ b/test/e2e/mnist_pytorch_appwrapper_test.go
@@ -145,6 +145,9 @@ func runMnistPyTorchAppWrapper(t *testing.T, accelerator string, numberOfGpus in
 		},
 	}
 
+	raw := Raw(test, job)
+	raw = RemoveCreationTimestamp(test, raw)
+
 	// Create an AppWrapper resource
 	aw := &mcadv1beta2.AppWrapper{
 		TypeMeta: metav1.TypeMeta{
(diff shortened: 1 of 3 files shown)


test/e2e/mnist_pytorch_appwrapper_test.go
```diff
@@ -145,6 +147,10 @@ func runMnistPyTorchAppWrapper(t *testing.T, accelerator string, numberOfGpus in
 		},
 	}
 
+	raw, _ := json.Marshal(job)
+	// Workaround to remove creationTimestamp which is brought by json marshaller, this field breaks Kueue reconciliation
+	patchedRaw := strings.ReplaceAll(string(raw), `"metadata":{"creationTimestamp":null},`, "")
```
astefanutti, line 152: Could this be encapsulated into the `Raw` helper function?

sutaakar: I can add it there as an option

sutaakar: Added the replace into dedicated helper function.
I didn't want to add it to the `Raw` as the manipulation of the content needs to be done on a struct pointer, so in the end I have manipulated with `RawExtension` anyway. In such case `RawExtension` can be modified outside of `Raw`.

PR #642, opened 2024-12-16
add e2e tests for AppWrappers with Jobs, Deployments, and Services

dgrove-oss:
Follow up from #640  -- added two additional e2e tests for AppWrappers.  The Deployment+Service one would have failed before the fix.

### DIFF
--- a/test/e2e/deployment_appwrapper_test.go
+++ b/test/e2e/deployment_appwrapper_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2023.
+Copyright 2024.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


test/e2e/deployment_appwrapper_test.go
```diff
@@ -0,0 +1,175 @@
...
+	"sigs.k8s.io/kueue/apis/kueue/v1beta1"
+)
+
+func TestDeploymentAppWrapperCpu(t *testing.T) {
+	runDeploymentAppWrapper(t)
+}
+
+func TestDeploymentAppWrapperGpu(t *testing.T) {
+	runDeploymentAppWrapper(t)
+}
```
sutaakar, line 42: Just a note, in other tests the CPU/GPU test split is done as GPU test has adjusted configuration.
If you don't have any specific GPU related code feel free to keep only one test function.

dgrove-oss: Thanks.  Done.

dgrove-oss:
Test cleaned up and ready for review.
On my cluster with the codeflare operator from RHOAI 2.16 installed `TestDeploymentAppWrapper` fails (as expected because it doesn't have the fix from #640).

varshaprasad96 (approved):
/lgtm

PR #650, opened 2025-01-14
Add automated update step for component metadata file in workflow

oksanabaza:
# Issue link

https://issues.redhat.com/browse/RHOAIENG-16074
# What changes have been made

Added automated update step for component metadata file in workflow
# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/.github/workflows/project-codeflare-release.yml
+++ b/.github/workflows/project-codeflare-release.yml
@@ -165,3 +165,20 @@ jobs:
           echo "Kueue release with version ${{ github.event.inputs.kueue-version }} does not exist. Please select an existing version."
           exit 1
         fi
+
+  generate-component-metadata:
+    needs: [check-appwrapper-version, check-kuberay-version, check-kueue-version]
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Generate component_metadata.yaml
+      run: |
+        cat <<EOL > config/component_metadata.yaml
+        releases:
+          - name: CodeFlare Operator
+            version: ${{ github.event.inputs.operator-version }}
+            repoUrl: https://github.com/project-codeflare/codeflare-operator
+        EOL
+
+    - name: Verify generated file
+      run: cat config/component_metadata.yaml


.github/workflows/project-codeflare-release.yml
```diff
@@ -165,3 +165,29 @@ jobs:
...
         fi
+
+  generate-component-metadata:
+    needs: [check-appwrapper-version, check-kuberay-version, check-kueue-version]
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Generate component_metadata.yaml
+      run: |
+        cat <<EOL > config/component_metadata.yaml
```
astefanutti, line 176: I fail to see how this can update the files in the corresponding repositories. Am I missing it?

Also it does not cover the training operator right?

.github/workflows/project-codeflare-release.yml
```diff
@@ -165,3 +165,29 @@ jobs:
...
+        releases:
```
Fiona-Waters, line 177: As far as I know we should just include the components that are referenced in the DSC resource, so codeflare, ray, kueue and training operator. We need to update separate files in separate repos for each.
We need to add training operator but as it is not GA we don't want to include it in the stack compatibility matrix in the readme yet. 
I wonder if a separate workflow that is called in the release workflow would work better.

sutaakar:
I think that the metadata upgrade should be part of component release/upgrade workflow, as that is the point when component is changed.
For Ray/Training operator/Kueue the release/upgrade is independent action from CodeFlare release. So I think that as part of CodeFlare release we can upgrade just version of CFO(AppWrapper?)/SDK.

.github/workflows/project-codeflare-release.yml
oksanabaza: There is currently no way to fully automate updating the files in the corresponding repositories. The version of the component can only be updated during the rebase/sync process. The purpose of this PR is to autogenerate the component_metadata file as part of the release process, ensuring all versions are accurately captured from the input fields before the release.

I haven’t included the training operator because it is not part of our release parameters

sutaakar (commented):
/lgtm

Fiona-Waters (commented):
/lgtm
@sutaakar @astefanutti I think we can merge this now

PR #658, opened 2025-01-28
update codeflare common to pick up fix for nvidia gpu setup

dgrove-oss:


sutaakar (approved):

sutaakar:
Just a note - the `codeflare-comon` fix in GitHub action is picked up automatically as the workflow refer to main branch - https://github.com/project-codeflare/codeflare-operator/blob/main/.github/workflows/e2e_tests.yaml#L42

PR #669, opened 2025-03-06
Add constant namespaces for RayDashboard tests

ChristianZaccaria:
## In relation to
Running e2e tests locally.

## Changes
Adding constant namespace names to tests that require to add the ray dashboard URL to `/etc/hosts`. This way, we can run the e2e tests locally by pre-setting the host before running the tests.

## Requires
PR: https://github.com/project-codeflare/codeflare-common/pull/93

### DIFF
--- a/test/e2e/mnist_rayjob_raycluster_test.go
+++ b/test/e2e/mnist_rayjob_raycluster_test.go
@@ -50,7 +50,7 @@ func TestMnistRayJobRayClusterGpu(t *testing.T) {
 func runMnistRayJobRayCluster(t *testing.T, accelerator string, numberOfGpus int) {
 	test := With(t)
 
-	// Create a namespace
+	// Create a static namespace to ensure a consistent Ray Dashboard hostname entry in /etc/hosts before executing the test.
 	namespace := test.NewTestNamespace(WithNamespaceName("test-ns-1"))
 
 	// Create Kueue resources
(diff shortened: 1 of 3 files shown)


test/e2e/mnist_rayjob_raycluster_test.go
```diff
@@ -51,7 +51,7 @@ func runMnistRayJobRayCluster(t *testing.T, accelerator string, numberOfGpus int
 	test := With(t)
 
 	// Create a namespace
-	namespace := test.NewTestNamespace()
+	namespace := test.NewTestNamespace(WithNamespaceName("test-ns-1"))
```
Bobbins228, line 54: Can you add a comment for future ref for why this test in particular is set to have a static name. 

Something like 
```
# Testing namespace left static for adding the Ray Dashboard to /etc/hosts 
```

test/e2e/mnist_rayjob_raycluster_test.go
```diff
@@ -122,7 +122,7 @@ func runMnistRayJobRayClusterAppWrapper(t *testing.T, accelerator string, number
 	test := With(t)
 
 	// Create a namespace
-	namespace := test.NewTestNamespace()
+	namespace := test.NewTestNamespace(WithNamespaceName("test-ns-2"))
```
Bobbins228, line 125: Same here

sutaakar (approved):

PR #673, opened 2025-03-20
Bump AppWrapper version to 1.1.2 (Kueue 0.11 compatible)

dgrove-oss:
**This PR should not be merged until we are ready to upgrade Kueue to 0.11.x.**

This PR bumps the AppWrapper version to 1.1.1, which is built to work with Kueue 0.11.

The key change is that in Kueue 0.11, AppWrapper is a built-in integration. This enables the removal of all Kueue-specific code from the AppWrapper controller, which eliminates the tight coupling between AppWrapper and Kueue versions.


go.mod
```diff
@@ -4,76 +4,84 @@ go 1.23.0
...
-	sigs.k8s.io/kueue v0.10.1
+	k8s.io/utils v0.0.0-20241210054802-24370beab758
+	sigs.k8s.io/controller-runtime v0.20.3
+	sigs.k8s.io/kueue v0.11.0
 	sigs.k8s.io/yaml v1.4.0
 )
 
 replace k8s.io/client-go => k8s.io/client-go v0.31.4
 
+replace k8s.io/api => k8s.io/api v0.31.4
```
dgrove-oss, line 32: There are more replace directives here than I'd like, but it was expedient in the short term. Someone who has a better understanding of how to coordinate Kubernetes/OpenShift API versions may be able to do something less heavy handed.

go.mod
```diff
@@ -4,76 +4,84 @@ go 1.23.0
...
+
+replace k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.31.2
+
+replace k8s.io/apimachinery v0.32.3 => k8s.io/apimachinery v0.31.4
+
+replace sigs.k8s.io/controller-runtime v0.20.3 => sigs.k8s.io/controller-runtime v0.19.3
+
+replace sigs.k8s.io/kueue v0.11.0 => sigs.k8s.io/kueue v0.10.1
```
varshaprasad96, line 40: This seems weird. We usually have to maintain compatibility between k8s, Kueue, AppWrapper.
cc: @sutaakar

dgrove-oss: One thing to consider is that we use the KUEUE_VERSION in the Makefile for two things:
1. Specifying which version of Kueue should be deployed in the CI testing.
2. Specifying which version of Kueue's golang API the e2e test code depends on.

For this PR, it is important that the Kueue version of reason 1 is 0.11 (because AppWrapper 1.1 won't work with Kueue 0.10).  However, for 2 it would have been fine to leave Kueue at 0.10.1.  AppWrapper no longer imports Kueue at all, so the only reason we have it is for the APIs used in the e2e test code.  And those are much less sensitive to exact Kueue version.

WDYT about have two variables for the two use cases and allowing them to diverge?  I think it would make your downstreaming simpler.

sutaakar: It is fine for me to separate the Kueue versions.
If I get it right it should be enough to remove https://github.com/project-codeflare/codeflare-operator/blob/c03f1b503a10e51c0c8a4d7c73157096a26062b3/Makefile#L163.

dgrove-oss: I pushed a change that removes line 163 as suggested.  I think this is the right thing to do.

sutaakar: Looks good

PR #676, opened 2025-03-27
only use the applicationsNamespace in openshift

szaher:
the codeflare operator checks for kuberay namespaces to apply network policies so it uses opendatahub DSCInitialization CRD which isn't available on vanilla k8s clusters so added a condition to check for DSCInitialization only when running on openshift

This fixes the gatejobs running the codeflare-sdk repo as the e2e test jobs are failing because DSCInitialization isn't available on kind cluster that is being used for testing.

# Issue link

# What changes have been made

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/pkg/controllers/raycluster_controller.go
+++ b/pkg/controllers/raycluster_controller.go
@@ -269,18 +269,21 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	// Locate the KubeRay operator deployment:
 	// - First try to get the ODH / RHOAI application namespace from the DSCInitialization
 	// - Or fallback to the well-known defaults
+	// add check if running on openshift or vanilla kubernetes
 	var kubeRayNamespaces []string
-	dsci := &dsciv1.DSCInitialization{}
-	err := r.Client.Get(ctx, client.ObjectKey{Name: "default-dsci"}, dsci)
-	if errors.IsNotFound(err) {
-		kubeRayNamespaces = []string{"opendatahub", "redhat-ods-applications"}
-	} else if err != nil {
-		return ctrl.Result{}, err
-	} else {
-		kubeRayNamespaces = []string{dsci.Spec.ApplicationsNamespace}
+	kubeRayNamespaces = []string{"opendatahub", "redhat-ods-applications"}
+
+	if r.IsOpenShift {
+		dsci := &dsciv1.DSCInitialization{}
+		err := r.Client.Get(ctx, client.ObjectKey{Name: "default-dsci"}, dsci)
+		if err != nil {
+			return ctrl.Result{}, err
+		} else {
+			kubeRayNamespaces = []string{dsci.Spec.ApplicationsNamespace}
+		}
 	}
 
-	_, err = r.kubeClient.NetworkingV1().NetworkPolicies(cluster.Namespace).Apply(ctx, desiredHeadNetworkPolicy(cluster, r.Config, kubeRayNamespaces), metav1.ApplyOptions{FieldManager: controllerName, Force: true})
+	_, err := r.kubeClient.NetworkingV1().NetworkPolicies(cluster.Namespace).Apply(ctx, desiredHeadNetworkPolicy(cluster, r.Config, kubeRayNamespaces), metav1.ApplyOptions{FieldManager: controllerName, Force: true})
 	if err != nil {
 		logger.Error(err, "Failed to update NetworkPolicy")
 	}


pkg/controllers/raycluster_controller.go
```diff
@@ -269,18 +269,21 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 	var kubeRayNamespaces []string
-	dsci := &dsciv1.DSCInitialization{}
-	err := r.Client.Get(ctx, client.ObjectKey{Name: "default-dsci"}, dsci)
-	if errors.IsNotFound(err) {
-		kubeRayNamespaces = []string{"opendatahub", "redhat-ods-applications"}
-	} else if err != nil {
-		return ctrl.Result{}, err
-	} else {
-		kubeRayNamespaces = []string{dsci.Spec.ApplicationsNamespace}
+	kubeRayNamespaces = []string{"opendatahub", "redhat-ods-applications"}
```
kryanbeane, line 274: @szaher We're still passing these namespaces (which I assume are openshift specific) to the network policy creation below. I may be wrong but on a native k8s installation, kuberay won't be installed in either of those namespaces. The OpenShift specific [kustomization.yaml](https://github.com/opendatahub-io/kuberay/blob/664807988d12d47c588cb706250027115b4fe6bd/ray-operator/config/openshift/kustomization.yaml#L4) points to opendatahub by default but the deployments namespace by default I think is just `ray-system`.  So wouldn't this still set up those network policies to point to ODH/RHOAI namespaces?

kryanbeane (approved):

pkg/controllers/raycluster_controller.go
```diff
@@ -269,18 +269,21 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+
+	if r.IsOpenShift {
+		dsci := &dsciv1.DSCInitialization{}
+		err := r.Client.Get(ctx, client.ObjectKey{Name: "default-dsci"}, dsci)
+		if err != nil {
+			return ctrl.Result{}, err
+		} else {
+			kubeRayNamespaces = []string{dsci.Spec.ApplicationsNamespace}
+		}
```
eoinfennessy, line 283:
```suggestion
var kubeRayNamespaces []string
	if r.IsOpenShift {
		dsci := &dsciv1.DSCInitialization{}
		if err := r.Client.Get(ctx, client.ObjectKey{Name: "default-dsci"}, dsci); err != nil {
			return ctrl.Result{}, err
		}
		kubeRayNamespaces = []string{dsci.Spec.ApplicationsNamespace}
	} else {
		kubeRayNamespaces = []string{"opendatahub", "redhat-ods-applications"}
```

chipspeak (approved):
LGTM Thanks, Saad!

eoinfennessy (commented):
Nit: Small readability change avoiding mutation of `kubeRayNamespaces` and removing unneeded `else` block after early return.

Otherwise LGTM.

pkg/controllers/raycluster_controller.go
szaher: Thanks @eoinfennessy actually there is a missing condition that needs to be handled. if the dsci is not found error we use the fall back namespaces if it's available we use it.
otherwise if it's not openshift, I will use the cluster namespace

pkg/controllers/raycluster_controller.go
szaher: Thanks @kryanbeane, I think the kuberay operator namespace is added via https://github.com/project-codeflare/codeflare-operator/blob/main/pkg/controllers/raycluster_controller.go#L553

pkg/controllers/raycluster_controller.go
```diff
@@ -271,16 +271,19 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-	kubeRayNamespaces = []string{"opendatahub", "redhat-ods-applications"}
+	kubeRayNamespaces = []string{cluster.Namespace}
 
 	if r.IsOpenShift {
 		dsci := &dsciv1.DSCInitialization{}
+
 		err := r.Client.Get(ctx, client.ObjectKey{Name: "default-dsci"}, dsci)
-		if err != nil {
+		if errors.IsNotFound(err) {
+			kubeRayNamespaces = []string{"opendatahub", "redhat-ods-applications"}
```
eoinfennessy, line 281: This variable always gets changed on line 285, so `kubeRayNamespaces` can never be set to `[]string{"opendatahub", "redhat-ods-applications"}`.

szaher: good catch, you're right 👍 I'll fix it now

pkg/controllers/raycluster_controller.go
```diff
@@ -269,18 +269,25 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 	var kubeRayNamespaces []string
-	dsci := &dsciv1.DSCInitialization{}
-	err := r.Client.Get(ctx, client.ObjectKey{Name: "default-dsci"}, dsci)
-	if errors.IsNotFound(err) {
-		kubeRayNamespaces = []string{"opendatahub", "redhat-ods-applications"}
-	} else if err != nil {
-		return ctrl.Result{}, err
-	} else {
-		kubeRayNamespaces = []string{dsci.Spec.ApplicationsNamespace}
+	kubeRayNamespaces = []string{cluster.Namespace}
```
sutaakar, line 274: Why is RayCluster namespace used here for KubeRay namespace?

sutaakar: This code implicates that KubeRay operator and RayCluster CR are installed in the same namespace.
Which IMHO is not correct.

szaher: @sutaakar are you suggesting that we add a new config item to the `KubeRayConfiguration` to allow users to use their own namespace for ray operator if not default one or is it already available? 

checked here https://github.com/project-codeflare/codeflare-operator/blob/main/pkg/config/config.go#L47 has no `namespace` in the config

sutaakar: fine for me to adjust it later

Issue #684, opened 2025-05-14
unit tests failing


laurafitzgerald:
### Describe the Bug

Unit tests are failing because Route object is not available.

### Codeflare Stack Component Versions
prs to main

### Steps to Reproduce the Bug

Create any pr e.g. https://github.com/project-codeflare/codeflare-operator/pull/683

### What Have You Already Tried to Debug the Issue?

Investigated the logs of https://github.com/project-codeflare/codeflare-operator/actions/runs/15018943191/job/42203318816

This appears to be the offending error `  2025-05-14T11:01:43Z	ERROR	controller-runtime.source.EventHandler	if kind is a CRD, it should be installed before calling Start	{"kind": "Route.route.openshift.io", "error": "no matches for kind \"Route\" in version \"route.openshift.io/v1\""}

### Expected Behavior

Unit tests should run with Route object available. 
Also tests should fail early if the files are missing

### Screenshots, Console Output, Logs, etc.

Add screenshots of UIs (like dashboards), etc. that help explain the issue.

### Affected Releases

latest main

PR #685, opened 2025-05-14
update route link as previous file was removed

laurafitzgerald:
# Issue link
https://github.com/project-codeflare/codeflare-operator/issues/684

# What changes have been made
update file source for Route CRD.
Also fail early if the http request returns a 404 as it did in this case. 

# Verification steps
Tests should pass

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


kryanbeane (approved):
good spot :)

chipspeak (approved):
Thanks Laura!

